PR #20, opened 2023-11-23
Added WhitespacePreTokenizer

jkrukowski:
In this PR:
- added `WhitespacePreTokenizer`
- added pretokenizer tests
- removed `public` from internal class methods

### DIFF
--- a/Sources/Tokenizers/PreTokenizer.swift
+++ b/Sources/Tokenizers/PreTokenizer.swift
@@ -55,7 +55,7 @@ struct PreTokenizerFactory {
         case .Digits: return DigitsPreTokenizer(config: config)
         case .Split: return SplitPreTokenizer(config: config)
         case .Whitespace: return WhitespacePreTokenizer(config: config)
-        default       : fatalError("Unsupported PreTokenizer type: \(typeName)")
+        default: fatalError("Unsupported PreTokenizer type: \(typeName)")
         }
     }
 }
--- a/Sources/Tokenizers/PreTokenizer.swift
+++ b/Sources/Tokenizers/PreTokenizer.swift
@@ -38,6 +38,7 @@ enum PreTokenizerType: String {
     case Punctuation
     case Digits
     case Split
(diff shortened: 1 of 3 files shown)


Tests/PreTokenizerTests/PreTokenizerTests.swift
```diff
@@ -0,0 +1,153 @@
...
+//  PreTokenizerTests.swift
+//
+//  Created by Jan Krukowski on 23/11/2023.
+//
+
+import XCTest
+import Hub
+@testable import Tokenizers
+
+class PreTokenizerTests: XCTestCase {
```
pcuenca, line 11: This alone is extremely useful, thanks a lot!

Sources/Tokenizers/PreTokenizer.swift
```diff
@@ -53,6 +54,7 @@ struct PreTokenizerFactory {
         case .Punctuation: return PunctuationPreTokenizer(config: config)
         case .Digits: return DigitsPreTokenizer(config: config)
         case .Split: return SplitPreTokenizer(config: config)
+        case .Whitespace: return WhitespacePreTokenizer(config: config)
         default       : fatalError("Unsupported PreTokenizer type: \(typeName)")
```
pcuenca, line 58:
```suggestion
        case .Whitespace: return WhitespacePreTokenizer(config: config)
        default: fatalError("Unsupported PreTokenizer type: \(typeName)")
```

Can we take the opportunity to remove this old alignment spacing that is no longer up to date? 🙏

pcuenca (approved):
Very clean and complete, thank you very much! 🙌

I just suggested an unrelated nit in a line adjacent to yours.

Sources/Tokenizers/PreTokenizer.swift
jkrukowski: done

pcuenca (approved):
Thanks again! 🙌

PR #21, opened 2023-11-25
Use the new top-k implementation in Accelerate

jkrukowski:
In this PR:
- added new top-k implementation using `Accelerate.BNNS.applyTopK` method
- added tests


Sources/TensorUtils/Math.swift
```diff
@@ -76,27 +75,38 @@ public struct Math {
...
+        try? Accelerate.BNNS.applyTopK(
+            k: k,
+            input: arrDescriptor,
+            bestValues: bestValues,
+            bestIndices: bestIndices,
+            axis: 0,
+            batchSize: 1,
+            filterParameters: nil
+        )
+        let distances = bestValues.data!.withMemoryRebound(to: Float.self, capacity: k) { ptr in
```
pcuenca, line 103: Is there anything sensible to do if the operation fails rather than crashing? I'm fine merging as is, just wandering if you have any ideas that don't complicate things too much.

Tests/TensorUtilsTests/TensorUtilsTests.swift
```diff
@@ -0,0 +1,100 @@
...
+        let lhsEvaluated = try expression1()
+        let rhsEvaluated = try expression2()
+        XCTAssertEqual(lhsEvaluated.count, rhsEvaluated.count, file: file, line: line)
+        for (lhs, rhs) in zip(lhsEvaluated, rhsEvaluated) {
+            XCTAssertEqual(lhs, rhs, accuracy: accuracy, file: file, line: line)
+        }
+    } catch {
+        XCTFail("Unexpected error: \(error)", file: file, line: line)
+    }
+}
```
pcuenca, line 100: Nice!

Sources/TensorUtils/Math.swift
jkrukowski: I'm ok with merging it as is for now as well. Documentation is rather silent about the error being thrown, so I suspect that the swift function definition might be autogenerated from some lower level C code. Making this function throwable might just overcomplicate it too much with not much of a gain.

jkrukowski: @pcuenca  do you think it'd be better to use `try!` here?

pcuenca: Nah, I guess it will fail in the following line anyway.

pcuenca: Actually, yes, let's do that just in case.

jkrukowski: done, forgot to call `deallocate` on `BNNSNDArrayDescriptor` so added it as well

pcuenca (approved):

pcuenca:
Amazing work, thanks again!

PR #23, opened 2023-11-30
Add top-p sampling

jkrukowski:
In this PR
- added top-p sampling function based on https://gist.github.com/thomwolf/1a5a29f6962089e871b94cbd09daf317
- added tests

I've compared 2 different implementations here https://github.com/jkrukowski/topp -- looks like using Accelerate to compute a cumulative sum gives it a seed boost.

### DIFF
--- a/Sources/Generation/Generation.swift
+++ b/Sources/Generation/Generation.swift
@@ -67,8 +67,11 @@ public extension Generation {
             if config.topK > 0 {
                 let topK = Math.topK(arr: logits, k: config.topK)
                 nextToken = Math.sample(indexes: topK.indexes, probs: topK.probs)
+            } else if config.topP < 1.0 {
+                let topP = Math.topP(arr: logits, p: Float(config.topP))
+                nextToken = Math.sample(indexes: topP.indexes, probs: topP.probs)
             } else {
-                fatalError("topP not implemented yet")
+                fatalError("not implemented yet")
             }
             
             if nextToken == config.eosTokenId { break }
--- a/Sources/TensorUtils/Math.swift
+++ b/Sources/TensorUtils/Math.swift
@@ -68,6 +68,48 @@ public struct Math {
         let ptr = UnsafeMutablePointer<Float32>(OpaquePointer(multiArray.dataPointer))
         return Math.argmax32(ptr, count: multiArray.count)
     }
+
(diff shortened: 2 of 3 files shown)


pcuenca (commented):
This is fantastic, @jkrukowski!

Let me try it out and check the implementation against the current one in `transformers` in case there are some details that could be incorporated, but this looks great already.

As a side comment, we could potentially implement the costly `cumsum` operation in Core ML as part of the model conversion, or using a Core ML pipeline. But using `Accelerate` should be more than enough for now!

Sources/Generation/Generation.swift
```diff
@@ -67,8 +67,11 @@ public extension Generation {
             if config.topK > 0 {
                 let topK = Math.topK(arr: logits, k: config.topK)
                 nextToken = Math.sample(indexes: topK.indexes, probs: topK.probs)
+            } else if config.topP < 1.0 {
```
pcuenca, line 70: If my understanding of this is correct, top-k can coexist with top-p: https://github.com/huggingface/transformers/blob/42017d82baa083da2bee3055fdac80c81ee97b8a/src/transformers/generation/utils.py#L805-L808

However, it could make sense to merge this PR now and making them coexist in a future one. What do you think?

Sources/Generation/Generation.swift
```diff
@@ -67,8 +67,11 @@ public extension Generation {
...
+                let topP = Math.topP(arr: logits, p: Float(config.topP))
+                nextToken = Math.sample(indexes: topP.indexes, probs: topP.probs)
             } else {
-                fatalError("topP not implemented yet")
+                fatalError("not implemented yet")
```
pcuenca, line 74: If we make top-k compatible with top-p, we'd do a single `sample` call on the selected tokens and remove this `fatalError`.

pcuenca (approved):
I tested it and it works fine! As expected, it's a bit slow but we can try to optimize later. Additionally, top-k and top-p could potentially coexist as pointed out below, but we can also handle that in a new PR unless you want to tackle it now :)

Sources/Generation/Generation.swift
jkrukowski: I'd say let's merge it now, seems logical to create a separate PR with a common interface to these two

PR #25, opened 2023-12-04
Support `.mlmodelc` in CLI

pcuenca:
This allows caching of models specialized for the current hardware. When using `.mlpackage`, the model is first compiled to `.mlmodelc` (fast), and then the `.mlmodelc` is specialized for the current hardware. In the first case, the specialized version is cached; in the second it is discarded.

See https://apple.github.io/coremltools/docs-guides/source/model-prediction.html#why-use-a-compiled-model


pcuenca:
Related: https://github.com/huggingface/swift-chat/pull/14

PR #27, opened 2023-12-09
Allow top-k and top-p to coexist

jkrukowski:
This PR is based on the comment https://github.com/huggingface/swift-transformers/pull/23#discussion_r1414318147

@pcuenca please let me know if that's what you had in mind.

This PR:
- adds `LogitsWrapper` based on https://github.com/huggingface/transformers/blob/42017d82baa083da2bee3055fdac80c81ee97b8a/src/transformers/generation/logits_process.py#L62
- adds `LogitsProcessor` based on https://github.com/huggingface/transformers/blob/42017d82baa083da2bee3055fdac80c81ee97b8a/src/transformers/generation/logits_process.py#L73
- changes generation to use `LogitsProcessor`
- adds tests


Sources/Generation/Generation.swift
```diff
@@ -102,4 +89,18 @@ public extension Generation {
         
         return tokenizer.decode(tokens: output)
     }
+
+    private func logitsWrappers(config: GenerationConfig) -> [any LogitsWarper] {
```
pcuenca, line 93:
```suggestion
    private func logitsWarpers(config: GenerationConfig) -> [any LogitsWarper] {
```

Let's make sure we always use the right nomenclature `LogitsWarper` in all classes and variable names :)

Sources/Generation/Generation.swift
```diff
@@ -102,4 +89,18 @@ public extension Generation {
...
+        if config.temperature > 0 && config.temperature != 1 {
+            logitsWrappers.append(TemperatureLogitsWraper(temperature: Float(config.temperature)))
+        }
+        if config.topK > 0 {
+            logitsWrappers.append(TopKLogitsWarper(k: config.topK))
+        }
+        if config.topP < 1.0 {
+            logitsWrappers.append(TopPLogitsWarper(p: Float(config.topP)))
+        }
+        return logitsWrappers
```
pcuenca, line 104:
```suggestion
        var logitsWarpers = [any LogitsWarper]()
        if config.temperature > 0 && config.temperature != 1 {
            logitsWarpers.append(TemperatureLogitsWarper(temperature: Float(config.temperature)))
        }
        if config.topK > 0 {
            logitsWarpers.append(TopKLogitsWarper(k: config.topK))
        }
        if config.topP < 1.0 {
            logitsWarpers.append(TopPLogitsWarper(p: Float(config.topP)))
        }
        return logitsWarpers
```

Sources/TensorUtils/LogitsWarper/LogitsProcessor.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+public struct LogitsProcessor {
+    public var logitsWrappers: [any LogitsWarper]
```
pcuenca, line 4: same here in all members :)

Sources/TensorUtils/LogitsWarper/TemperatureLogitsWarper.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public struct TemperatureLogitsWraper: LogitsWarper {
```
pcuenca, line 3:
```suggestion
public struct TemperatureLogitsWarper: LogitsWarper {
```

Sources/TensorUtils/LogitsWarper/LogitsWrapper.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+/// Protocol for all logit warpers that can be applied during generation
+public protocol LogitsWarper {
+    func callAsFunction(_ arr: [Float]) -> (indexes: [Int], logits: [Float])
```
pcuenca, line 5: This is great, but out of personal preference I'd also include a method with a semantically descriptive name of what it does, and then use `callAsFunction` for convenience.

So in this case we'd have a method called `warp`, and then we'd have a default implementation for `callAsFunction` that would simply invoke `warp`. See for example how it's done here: https://github.com/huggingface/swift-transformers/blob/fa2522162afa259fe87246ffa41dec9e87e1f195/Sources/Tokenizers/Decoder.swift#L11-L22

Sources/TensorUtils/LogitsWarper/TemperatureLogitsWarper.swift
```diff
@@ -0,0 +1,14 @@
...
+    public var temperature: Float
+    
+    public init(temperature: Float) {
+        self.temperature = temperature
+    }
+
+    public func callAsFunction(_ arr: [Float]) -> (indexes: [Int], logits: [Float]) {
```
pcuenca, line 10: This would be renamed `warp` if we apply the change mentioned above.

Sources/TensorUtils/LogitsWarper/TopKLogitsWarper.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+import Accelerate
+
+/// Top-K.
+/// Select the k most-probable elements indices from `arr`
```
pcuenca, line 5:
```suggestion
/// Select the k most-probable element indices from `arr`
```

Sources/TensorUtils/LogitsWarper/TopKLogitsWarper.swift
```diff
@@ -0,0 +1,58 @@
...
+/// and return both the indices (from the original array)
+/// and their probabilities.
+public struct TopKLogitsWarper: LogitsWarper {
+    public var k: Int
+    
+    public init(k: Int) {
+        self.k = k
+    }
+
+    public func callAsFunction(_ arr: [Float]) -> (indexes: [Int], logits: [Float]) {
```
pcuenca, line 15: Same here, this would be the `warp` implementation.

Sources/TensorUtils/LogitsWarper/TopPLogitsWarper.swift
```diff
@@ -0,0 +1,37 @@
...
+/// Select the smallest set of elements whose cumulative probability exceeds the probability `p`.
+/// Based on https://gist.github.com/thomwolf/1a5a29f6962089e871b94cbd09daf317
+public struct TopPLogitsWarper: LogitsWarper {
+    public var p: Float
+
+    public init(p: Float) {
+        self.p = p
+    }
+
+    public func callAsFunction(_ arr: [Float]) -> (indexes: [Int], logits: [Float]) {
```
pcuenca, line 13: Same, this would become `warp`

Tests/TensorUtilsTests/LogitsWraperTests.swift
pcuenca, line 1: Thanks a lot for adding tests too! 🙌

pcuenca (commented):
Great job, @jkrukowski, sorry for taking long to review 🙏 

I think the PR is in very good shape and can be merged quickly. I like that it follows the `transformers` pattern and is super clean and easy to follow. I only have two comments:

- We should always use the nomenclature `warper` in all file, class, and variable names.
- I think it's a bit more expressive if the `LogitsWarper` protocol has a method called `warp`, and a default `callAsFunction` method that simply calls `warp`. We'd use the implicit call most of the time for symmetry with transformers, but users may also choose to call `warp` if they prefer, and the semantics would be clear.

Thanks a lot for another great contribution!

jkrukowski:
@pcuenca applied the requested changes, please take a 2nd look

pcuenca (approved):
Thanks a lot!

PR #29, opened 2023-12-20
Whisper Tokenizer and `added_tokens`

pcuenca:


pcuenca:
Tokenizer tests pass, so I'll merge for simplicity. We need to:

- Better test `WhisperTokenizer` in the wild
- Handle `specialTokens` appropriately, they are currently ignored. This affects all tokenizers.

PR #32, opened 2023-12-26
Public token <-> id conversion + Unknown Tokens

pcuenca:
I used an API similar to transformers.
To do (in new PRs, potentially):

- Some tests
- Use unknown token id
- Handle special tokens

cc @ZachNagengast, feel free to review!

### DIFF
--- a/Sources/Tokenizers/BPETokenizer.swift
+++ b/Sources/Tokenizers/BPETokenizer.swift
@@ -38,6 +38,9 @@ class BPETokenizer: Tokenizer {
     private let addedTokens: Set<String>
     private let specialTokens: [String: Int]
     
+    public let unknownToken: String
+    public let unknownTokenId: Int
+    
     private let preTokenizer: PreTokenizer?
     private let normalizer: Normalizer?
     private let postProcessor: PostProcessor?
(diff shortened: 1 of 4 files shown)


ZachNagengast (approved):
Great addition - one idea regarding unk token, there could be a parameter in the calling interface for a default token/vocab to use if it's not found in the dictionaries, or these functions could even return an optional

Sources/Tokenizers/BPETokenizer.swift
```diff
@@ -38,13 +38,20 @@ class BPETokenizer: Tokenizer {
...
+    
     private let preTokenizer: PreTokenizer?
     private let normalizer: Normalizer?
     private let postProcessor: PostProcessor?
     private let decoder: Decoder?
     
     private let cleanUpTokenizationSpaces: Bool
 
+    class func unknownToken(from tokenizerConfig: Config) -> String? {
+        return tokenizerConfig.unkToken?.content?.stringValue ?? tokenizerConfig.unkToken?.stringValue
```
pcuenca, line 52: The first part is the modern way; the second expression appears in distilgpt2, for example.

Sources/Tokenizers/BPETokenizer.swift
```diff
@@ -38,13 +38,20 @@ class BPETokenizer: Tokenizer {
     private let addedTokens: Set<String>
     private let specialTokens: [String: Int]
     
+    public let unknownToken: String?
+    public let unknownTokenId: Int?
```
pcuenca, line 42: They can be nil now

pcuenca:
@ZachNagengast I further checked the Python code and brought over the relevant changes:

- The unknown token and unknown token id are now read from the tokenizer configuration.
- They can be `nil`. This happens, for example, in the Falcon tokenizer.

Tests now pass for the various tokenizers we are considering, and comparing against the well-known Python unknown token ids.

PR #41, opened 2024-01-23
BertTokenizer support Chinesechars

Designerxsh:
add CJK(Chinese, Japanese, Korean) support for BertTokenizer

copy from https://github.com/huggingface/transformers/blob/main/src/transformers/models/bert/tokenization_bert.py

### DIFF
--- a/Sources/Tokenizers/BertTokenizer.swift
+++ b/Sources/Tokenizers/BertTokenizer.swift
@@ -13,14 +13,18 @@ class BertTokenizer {
     private let basicTokenizer = BasicTokenizer()
     private let wordpieceTokenizer: WordpieceTokenizer
     private let maxLen = 512
+    private let tokenizeChineseChars: Bool
     
     private let vocab: [String: Int]
     private let ids_to_tokens: [Int: String]
     
-    init(vocab: [String: Int], merges: [String]?) {
+    init(vocab: [String: Int],
+         merges: [String]?,
+         tokenizeChineseChars: Bool = true) {
         self.vocab = vocab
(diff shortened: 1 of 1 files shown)


pcuenca:
Thanks a lot for the contribution! If you don't mind, maybe it makes sense to merge #40 first, and then try to reuse the Chinese codepoint detection [from there](https://github.com/huggingface/swift-transformers/pull/40#discussion_r1465408571).

pcuenca (commented):
Looks good! Can we maybe add a few tests?

Designerxsh:
> Thanks a lot for the contribution! If you don't mind, maybe it makes sense to merge #40 first, and then try to reuse the Chinese codepoint detection [from there](https://github.com/huggingface/swift-transformers/pull/40#discussion_r1465408571).

Ok, after #40 is merged, I will modify this part and use the method from the utility class to handle Chinese.

pcuenca:
@Designerxsh #40 has just been merged! :)

Designerxsh:
> @Designerxsh #40 has just been merged! :)

Hello, I've looked at the merge of #40 . The BertNormalizer in Normalizer.swift already supports CJK processing, so does the BertTokenizer still need separate support?

pcuenca:
Hi @Designerxsh! Ideally, it shouldn't. However, the `BertTokenizer` follows a different code path than the others, and currently everything is hardcoded - the normalizers are not loaded from the configuration file in this case. The best long-term solution would be to use the same mechanism we use for `BertTokenizer` as we do for the `Unigram` and `BPE` models. Would you be up to exploring that? Otherwise, we can still use this PR as a workaround until we tackle the final solution.

PR #45, opened 2024-02-05
BertTokenizer support Chinesechars

Designerxsh:
add CJK(Chinese, Japanese, Korean) support for BertTokenizer, related to #41


pcuenca (approved):
Looks good to me!

pcuenca:
For reference, Chinese chars tokenization is also [enabled by default in transformers](https://github.com/huggingface/transformers/blob/2da28c4b41bba23969a8afe97c3dfdcbc47a57dc/src/transformers/models/bert/tokenization_bert.py#L169C9-L169C26), and [this is the implementation](https://github.com/huggingface/transformers/blob/main/src/transformers/models/bert/tokenization_bert.py#L483).

pcuenca (approved):

PR #46, opened 2024-02-07
Simulator flag for background urlsessions 

ZachNagengast:
Background url sessions require entitlements to run properly, this allows simulators and in particular CI machines to run the code without entitlements by using default configs instead of background.


pcuenca (approved):
Thanks a lot!

PR #48, opened 2024-02-07
Add RobertaProcessing

shavit:


pcuenca (commented):
Thanks a lot! 🙌 

Looks good to me in principle. Could you please add some tests that compare results against the `tokenizers` reference implementation, as used in `transformers`?

In addition, if you followed any of the Python / Rust / JavaScript implementations, it'd be cool to include a link to the relevant portion in the comments, just as you did with the code from facebook.

shavit:
Yes, I didn't include the tests, because I had a question about the implementation of the tokens. 

The other libraries have different implementations and tests, but I included a link to the relevant part in `transformers`.  

Let me know if you think something is missing.

pcuenca (approved):
Thanks a lot!

Issue #44, opened 2024-02-02
TinyLlama


thejefflarson:
Hi!

I'm fiddling with [TinyLlama](https://huggingface.co/TinyLlama/TinyLlama-1.1B-Chat-v1.0) and trying to get it to work. First I ran:

`python -m exporters.coreml -m TinyLlama/TinyLlama-1.1B-Chat-v1.0 exported/ --feature "text-generation" --quantize float32`

(I also tried quantizing it to `float16`.)

To convert it to CoreML and then I created a singleton loader class to access it:

```swift
class LLM {
    static var shared: LanguageModel = {
        let url = Bundle.main.url(forResource: "TinyLlama", withExtension: "mlmodelc")!
        return try! LanguageModel.loadCompiled(url: url, computeUnits: .cpuAndGPU)
    }()
}
```

And finally I ran something like this:

```swift
do {
    let config = GenerationConfig(maxNewTokens: 256, doSample: true, temperature: 0.7, topK: 50, topP: 0.95)
    let output = try await LLM.shared.generate(config: config, prompt: prompt)
} catch {
    print("Error \(error)")
}
```

And it prints out gibberish, specifically a list of `<unk>` and `<\s>` (eos) tokens. Any thoughts on what I'm doing wrong here?

pcuenca:
Your process looks good to me! Did you verify that the tokenizer files are being found and loaded correctly? This depends on a custom metadata field added to the Core ML package, but it should be added by `exporters`.

I have been testing TinyLlama myself, but in a work-in-progress dev branch. I can replicate everything in `main` to see if there's something I'm missing now.

thejefflarson:
Hm,  I'm not sure, what is the name of the field? I'm not seeing a tokenizer field in the xcode list, is there a way to check?
<img width="410" alt="Screenshot 2024-02-09 at 11 41 59 AM" src="https://github.com/huggingface/swift-transformers/assets/55365/9e07b120-8409-41f4-926a-047ad597de39">

pcuenca:
Hi @thejefflarson, it's `co.huggingface.exporters.name` and it's correct – the tokenizer files will be downloaded from that location in the Hub.

I've been testing and it looks like it's a bug with top-p sampling. After conversion (using `float16`), the following works:

```bash
swift run transformers "I" --max-length 128 --compute-units cpuAndGPU /Users/pedro/Desktop/LLMs/TinyLlama-1.1B-Chat-v1.0-exporters.mlmodelc
```

[...]

PR #49, opened 2024-02-11
Fixed LogitsProcessor

jkrukowski:
Issue is described [here](https://github.com/huggingface/swift-transformers/issues/44)

When `LogitsProcessor` contains more than one instance of `LogitsWarper` wrong indexes are returned -- we're returning indexes of the processed array but we should return these indexes mapped to the result of the previous call of `LogitsProcessor`.

Fixed by changing the interface of `LogitsWarper` allowing to pass indexes along with logits and mapping the result indexes to the passed indexes.

### DIFF
--- a/Sources/TensorUtils/LogitsWarper/TopKLogitsWarper.swift
+++ b/Sources/TensorUtils/LogitsWarper/TopKLogitsWarper.swift
@@ -12,14 +12,14 @@ public struct TopKLogitsWarper: LogitsWarper {
         self.k = k
     }
 
-    public func warp(_ arr: [Float]) -> (indexes: [Int], logits: [Float]) {
-        guard !arr.isEmpty else {
+    public func warp(indexes: [Int], logits: [Float]) -> (indexes: [Int], logits: [Float]) {
+        guard !logits.isEmpty else {
             return (indexes: [], logits: [])
         }
-        let k = min(k, arr.count)
+        let k = min(k, logits.count)
         let arrDescriptor = BNNSNDArrayDescriptor.allocate(
-            initializingFrom: arr,
-            shape: .vector(arr.count)
+            initializingFrom: logits,
+            shape: .vector(logits.count)
         )
         defer {
             arrDescriptor.deallocate()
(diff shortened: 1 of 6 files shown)


Tests/TensorUtilsTests/LogitsWarperTests.swift
```diff
@@ -95,7 +109,14 @@ final class LogitsWarperTests: XCTestCase {
             logitsWarpers: [TopKLogitsWarper(k: 3), TopPLogitsWarper(p: 0.99)]
         )
         let result4 = processor4([2.0, 1.0, 3.0, -5.0, -23.0, 12.5])
-        XCTAssertEqual(result4.indexes, [0])
+        XCTAssertEqual(result4.indexes, [5])
```
jkrukowski, line 112: we had tests for this exact issue but we were asserting the wrong value

Sources/TensorUtils/LogitsWarper/TopKLogitsWarper.swift
```diff
@@ -47,12 +47,12 @@ public struct TopKLogitsWarper: LogitsWarper {
...
-        let distances = bestValues.data!.withMemoryRebound(to: Float.self, capacity: k) { ptr in
+        let topkLogits = bestValues.data!.withMemoryRebound(to: Float.self, capacity: k) { ptr in
             Array(UnsafeBufferPointer(start: ptr, count: k))
         }
-        let indices = bestIndices.data!.withMemoryRebound(to: Int32.self, capacity: k) { ptr in
+        let topkIndexes = bestIndices.data!.withMemoryRebound(to: Int32.self, capacity: k) { ptr in
             Array(UnsafeBufferPointer(start: ptr, count: k))
         }
-        return (indexes: indices.map { Int($0) }, logits: distances)
+        return (indexes: topkIndexes.map { indexes[Int($0)] }, logits: topkLogits)
```
jkrukowski, line 56: here we're mapping indexes

pcuenca (approved):
Thanks for looking into this so quick! I think we should settle on using the same nomenclature for indexes/indices, but let's fix that bug first.

PR #57, opened 2024-03-08
feat: Support custom endpoint

maiqingqiang:
Resolving the issue of some regions being unable to access https://huggingface.co

### DIFF
--- a/Sources/Hub/HubApi.swift
+++ b/Sources/Hub/HubApi.swift
@@ -8,14 +8,14 @@
 import Foundation
 
 public struct HubApi {
-    let endpoint = "https://huggingface.co/api"
     var downloadBase: URL
-    var hfToken: String? = nil
+    var hfToken: String?
+    var hfEndpoint: String
     
     public typealias RepoType = Hub.RepoType
     public typealias Repo = Hub.Repo
     
-    public init(downloadBase: URL? = nil, hfToken: String? = nil) {
+    public init(downloadBase: URL? = nil, hfToken: String? = nil, hfEndpoint: String = "https://huggingface.co") {
         if downloadBase == nil {
             let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
             self.downloadBase = documents.appending(component: "huggingface")
@@ -23,6 +23,7 @@ public struct HubApi {
             self.downloadBase = downloadBase!
         }
         self.hfToken = hfToken
+        self.hfEndpoint = hfEndpoint
     }
     
     static let shared = HubApi()
@@ -51,7 +52,7 @@ public extension HubApi {
         switch response.statusCode {
         case 200..<300: break
         case 400..<500: throw Hub.HubClientError.authorizationRequired
-        default       : throw Hub.HubClientError.httpStatusCode(response.statusCode)
+        default: throw Hub.HubClientError.httpStatusCode(response.statusCode)
         }
 
         return (data, response)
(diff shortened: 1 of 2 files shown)


Sources/Hub/HubApi.swift
```diff
@@ -8,21 +8,22 @@
 import Foundation
 
 public struct HubApi {
-    let endpoint = "https://huggingface.co/api"
     var downloadBase: URL
-    var hfToken: String? = nil
+    var hfToken: String?
+    var hfEndpoint: String
```
pcuenca, line 13: Why not keep `endpoint` here?

Sources/Hub/HubApi.swift
```diff
@@ -8,21 +8,22 @@
...
     
     public typealias RepoType = Hub.RepoType
     public typealias Repo = Hub.Repo
     
-    public init(downloadBase: URL? = nil, hfToken: String? = nil) {
+    public init(downloadBase: URL? = nil, hfToken: String? = nil, hfEndpoint: String = "https://huggingface.co") {
```
pcuenca, line 18: 👍

Sources/Hub/HubApi.swift
```diff
@@ -243,9 +245,8 @@ public extension Hub {
     }
 }
 
-public extension Array<String> {
+public extension [String] {
```
pcuenca, line 248: This is unrelated, but ok.

pcuenca (commented):
Thanks a lot! This is very useful, I only have one question about changing the property name. If anything, I'd remove `hf` from `hfToken` rather than adding it to `endpoint`.

Sources/Hub/HubApi.swift
maiqingqiang: This has been processed by the swift-format tool.

Sources/Hub/HubApi.swift
maiqingqiang: Because the token also has the "hf" prefix.😂

maiqingqiang:
> Thanks a lot! This is very useful, I only have one question about changing the property name. If anything, I'd remove `hf` from `hfToken` rather than adding it to `endpoint`.

I also think using `endpoint` is better. So, should I change `hfToken` to `token` on my end?

Sources/Hub/HubApi.swift
pcuenca: Yeah, I don't use it. I write what feels clearer in each case, I tend to use the short form for variables, and the spelled out format when working with the type itself, such as in extensions. But it's ok.

pcuenca:
> I also think using `endpoint` is better. So, should I change `hfToken` to `token` on my end?

I'd love that, but `hfToken` is already in the `init` signature, so we could break other people's code if we change it. Coming to think of it, let's keep `endpoint` for the same reason, if that's ok 🙏

maiqingqiang:
> > I also think using `endpoint` is better. So, should I change `hfToken` to `token` on my end?
> 
> I'd love that, but `hfToken` is already in the `init` signature, so we could break other people's code if we change it. Coming to think of it, let's keep `endpoint` for the same reason, if that's ok 🙏

Resolved

Sources/Hub/HubApi.swift
maiqingqiang: Can we introduce swift-format to unify the code style? Similar to https://github.com/nicklockwood/SwiftFormat?tab=readme-ov-file#git-pre-commit-hook or https://github.com/apple/swift-format with git pre commit hook

pcuenca (approved):
Thanks a lot!

PR #62, opened 2024-03-14
Propagate HubApi configuration

jkrukowski:
Propagating the `HubApi` configuration will allow the user to specify the download folder when using e.g. `AutoTokenizer.from` method.

Detailed description of the issue:

I'm trying to create a homebrew formula for [WhisperKit CLI](https://github.com/argmaxinc/WhisperKit) (PR is here: https://github.com/Homebrew/homebrew-core/pull/165912). As a part of the acceptance criteria, homebrew CI runs the tests on the formula in the sandbox environment. The tests are failing because the whisperkit cli is trying to download the tokenizer configuration to the destination to which it doesn't have access and this results in error:

```
Error: You don’t have permission to save the file “whisper-tiny” in the folder “openai”.
```

cc @ZachNagengast

### DIFF
--- a/Sources/Hub/Hub.swift
+++ b/Sources/Hub/Hub.swift
@@ -112,9 +112,15 @@ public class LanguageModelConfigurationFromHub {
 
     private var configPromise: Task<Configurations, Error>? = nil
 
-    public init(modelName: String) {
+    public init(
+        modelName: String,
+        hubApi: HubApi = HubApi()
+    ) {
         self.configPromise = Task.init {
-            return try await self.loadConfig(modelName: modelName)
+            return try await self.loadConfig(
+                modelName: modelName,
+                hubApi: hubApi
+            )
         }
     }
 
@@ -161,8 +167,10 @@ public class LanguageModelConfigurationFromHub {
         }
     }
 
-    func loadConfig(modelName: String, hfToken: String? = nil) async throws -> Configurations {
-        let hubApi = HubApi(hfToken: hfToken)
+    func loadConfig(
+        modelName: String,
+        hubApi: HubApi = HubApi()
+    ) async throws -> Configurations {
         let filesToDownload = ["config.json", "tokenizer_config.json", "tokenizer.json"]
         let repo = Hub.Repo(id: modelName)
         try await hubApi.snapshot(from: repo, matching: filesToDownload)
@@ -172,7 +180,11 @@ public class LanguageModelConfigurationFromHub {
         let tokenizerConfig = try? hubApi.configuration(from: "tokenizer_config.json", in: repo)
         let tokenizerVocab = try hubApi.configuration(from: "tokenizer.json", in: repo)
         
-        let configs = Configurations(modelConfig: modelConfig, tokenizerConfig: tokenizerConfig, tokenizerData: tokenizerVocab)
+        let configs = Configurations(
+            modelConfig: modelConfig,
+            tokenizerConfig: tokenizerConfig,
+            tokenizerData: tokenizerVocab
+        )
         return configs
     }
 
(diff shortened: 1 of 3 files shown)


Sources/Hub/Hub.swift
```diff
@@ -112,9 +112,15 @@ public class LanguageModelConfigurationFromHub {
 
     private var configPromise: Task<Configurations, Error>? = nil
 
-    public init(modelName: String) {
+    public init(
+        modelName: String,
+        hubApi: HubApi = HubApi()
```
pcuenca, line 117:
```suggestion
        hubApi: HubApi = HubApi.shared
```

maybe?

Sources/Hub/Hub.swift
```diff
@@ -161,8 +167,10 @@ public class LanguageModelConfigurationFromHub {
         }
     }
 
-    func loadConfig(modelName: String, hfToken: String? = nil) async throws -> Configurations {
-        let hubApi = HubApi(hfToken: hfToken)
+    func loadConfig(
+        modelName: String,
+        hubApi: HubApi = HubApi()
```
pcuenca, line 172:
```suggestion
        hubApi: HubApi = HubApi.shared
```

Same as before, unless you think it's better to instantiate

Sources/Hub/HubApi.swift
```diff
@@ -8,25 +8,29 @@
 import Foundation
 
 public struct HubApi {
-    var downloadBase: URL
-    var hfToken: String?
-    var endpoint: String
+    public var downloadBase: URL
+    public var hfToken: String?
+    public var endpoint: String
```
pcuenca, line 13:
```suggestion
    public let downloadBase: URL
    public let hfToken: String?
    public let endpoint: String
```

I think it's ok to make them public, but then I'd make them immutable

Sources/Hub/Hub.swift
```diff
@@ -112,9 +112,15 @@ public class LanguageModelConfigurationFromHub {
...
+    public init(
+        modelName: String,
+        hubApi: HubApi = HubApi()
+    ) {
         self.configPromise = Task.init {
-            return try await self.loadConfig(modelName: modelName)
+            return try await self.loadConfig(
+                modelName: modelName,
+                hubApi: hubApi
+            )
```
pcuenca, line 123: (Personally not a fan of multi-lines, especially if they're short, but let's not worry about that for the moment)

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -234,8 +234,11 @@ extension AutoTokenizer {
         return try PreTrainedTokenizer(tokenizerConfig: tokenizerConfig, tokenizerData: tokenizerData)
     }
 
-    public static func from(pretrained model: String) async throws -> Tokenizer {
-        let config = LanguageModelConfigurationFromHub(modelName: model)
+    public static func from(
+        pretrained model: String,
+        hubApi: HubApi = HubApi()
```
pcuenca, line 239:
```suggestion
        hubApi: HubApi = HubApi.shared
```

pcuenca (commented):
Not at all familiar with homebrew recipes, but I think the change is useful, thanks a lot!

Sources/Hub/Hub.swift
jkrukowski: make sense, changed

pcuenca (approved):
Thank you! Do you need this merged ASAP? Otherwise I'd wait for @ZachNagengast in case he sees a problem with it (tests pass on my end).

jkrukowski:
> Thank you! Do you need this merged ASAP? Otherwise I'd wait for @ZachNagengast in case he sees a problem with it (tests pass on my end).

sure, let's wait for @ZachNagengast

ZachNagengast:
Brilliant, this will help a ton with the pre-trained tokenizer downloads since normal model downloads allow the same feature.

jkrukowski:
@pcuenca would it be possible to tag a new release after this one is merged?

Issue #61, opened 2024-03-14
Download tasks getting rate limited by background configuration


ZachNagengast:
Following up on @davidkoski's #59  issue I did a bit of research and may have found the underlying issue based on this section in the docs: https://developer.apple.com/documentation/foundation/url_loading_system/downloading_files_in_the_background#3038438

> When the system resumes or relaunches your app, it uses a rate limiter to prevent abuse of background downloads. When your app starts a new download task while in the background, the task doesn't begin until the delay expires. The delay increases each time the system resumes or relaunches your app.
> 
> As a result, if your app starts a single background download, gets resumed when the download completes, and then starts a new download, it will greatly increase the delay. Instead, use a small number of background sessions — ideally just one — and use these sessions to start many download tasks at once. This allows the system to perform multiple downloads at once, and resume your app when they have completed.

This would explain the intermittent successes and hard-to-reproduce behavior, although this is still a hypothesis. This showed up for me when 1-2 downloads in a repo would succeed, but then it would hang up and not complete. I believe it is getting rate limited by the OS.

The reason it likely didn't show up in `swift-coreml-diffusers` is because that repo is downloading single zip files, and uses a static identifier that only creates one background session:

https://github.com/huggingface/swift-coreml-diffusers/blob/d041577b9f5e201baa3465bc60bc5d0a1cf7ed7f/Diffusion/Common/Downloader.swift#L32

Whereas now we are downloading an arbitrary number of separate files based on the glob search, and creating a new session for each file:

https://github.com/huggingface/swift-transformers/blob/95194e645743f9aebe9557bed045162701b70f9f/Sources/Hub/Downloader.swift#L36

I'd propose these two options that would be good to have feedback on:
1. Use a static identifier for background URLSession identifiers
2. Create an option to just use the default (foreground) URLSession configuration instead of the background configuration
    - Question: should this be the default for simplicity? Will changing this default break have the potential to break anyone's current usage?

I can have a PR ready with these changes, but curious what folks think about these findings cc @pcuenca @davidkoski @jkrukowski

davidkoski:
AFAIK background downloading is more interesting for iOS apps that want to download things past when the user is interacting with them.  If you want to download and wait for the download to finish I think a default download task would be more appropriate.

Perhaps the caller could indicate intent?

ZachNagengast:
[...]

Issue #59, opened 2024-03-13
download progress handler not being called


davidkoski:
tag 0.1.2

I see the progress handler for `hub.snapshot()` being called only at the beginning and end:

```
    let hub = HubApi()
    let repo = Hub.Repo(id: "mlx-community/starcoder2-3b-4bit")
    let modelFiles = ["config.json", "*.safetensors"]
    let modelDirectory = try await hub.snapshot(
        from: repo, matching: modelFiles) { progress in
            print(progress)
        }
```

gives:

```
<NSProgress: 0x600001d7e200> : Parent: 0x0 (portion: 0) / Fraction completed: 0.0000 / Completed: 0 of 2  
  <NSProgress: 0x600001d7ec80> : Parent: 0x600001d7e200 (portion: 1) / Fraction completed: 0.0000 / Completed: 0 of 100  

<NSProgress: 0x600001d7e200> : Parent: 0x0 (portion: 0) / Fraction completed: 1.0000 / Completed: 2 of 2  
```

I swear I have seen this get called for the download but I can't seem to reproduce it now. 

I can see:

```
extension Downloader: URLSessionDelegate, URLSessionDownloadDelegate {
    func urlSession(_: URLSession, downloadTask: URLSessionDownloadTask, didWriteData _: Int64, totalBytesWritten _: Int64, totalBytesExpectedToWrite _: Int64) {
```

get called and I see that this subscriber:

```
class Downloader: NSObject, ObservableObject {
...
    @discardableResult
    func waitUntilDone() throws -> URL {
        // It's either this, or stream the bytes ourselves (add to a buffer, save to disk, etc; boring and finicky)
        let semaphore = DispatchSemaphore(value: 0)
        stateSubscriber = downloadState.sink { state in
            switch state {
            case .completed: semaphore.signal()
            case .failed:    semaphore.signal()
            default:         break
            }
        }
```

Is receiving them.  I think the problem may be here:

```
public extension HubApi {
    func localRepoLocation(_ repo: Repo) -> URL {
        downloadBase.appending(component: repo.type.rawValue).appending(component: repo.id)
    }
    
    struct HubFileDownloader {
...
        @discardableResult
        func download(progressHandler: @escaping (Double) -> Void) async throws -> URL {
            guard !downloaded else { return destination }
            
            try prepareDestination()
            let downloader = Downloader(from: source, to: destination, using: hfToken)
            let downloadSubscriber = downloader.downloadState.sink { state in
[...]

Issue #39, opened 2024-01-17
[hub] Invalidate cache when contents of remote repo change


pcuenca:
We can also do file verification at the same time (something simple: check the file size)

FL33TW00D:
@greenrazer I think we discussed this before right?

pcuenca:
Yes. [This great PR](https://github.com/huggingface/swift-transformers/pull/155) very much goes in this direction and mimics how other libraries work.

I think we can close this issue now. We can also coordinate with @ardaatahan for next steps.

FL33TW00D:
Closing! thanks to @ardaatahan for the great contribution.

PR #64, opened 2024-03-14
Use default session config, with option for background session

ZachNagengast:
Prevents requests from getting rate limited by the OS, but still allows usage of background thread for developers that handle it appropriately.

See discussions here #61 #59  #46

### DIFF
--- a/Sources/Hub/Downloader.swift
+++ b/Sources/Hub/Downloader.swift
@@ -11,34 +11,42 @@ import Combine
 
 class Downloader: NSObject, ObservableObject {
     private(set) var destination: URL
-    
+
     enum DownloadState {
         case notStarted
         case downloading(Double)
         case completed(URL)
         case failed(Error)
     }
-    
+
     enum DownloadError: Error {
         case invalidDownloadLocation
         case unexpectedError
     }
-    
+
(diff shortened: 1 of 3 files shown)


Sources/Hub/Downloader.swift
```diff
@@ -83,31 +91,28 @@ class Downloader: NSObject, ObservableObject {
...
 
     func urlSession(_: URLSession, downloadTask _: URLSessionDownloadTask, didFinishDownloadingTo location: URL) {
-        guard FileManager.default.fileExists(atPath: location.path) else {
-            downloadState.value = .failed(DownloadError.invalidDownloadLocation)
-            return
-        }
         do {
-            try FileManager.default.moveItem(at: location, to: destination)
+            // If the downloaded file already exists on the filesystem, overwrite it
+            try FileManager.default.moveDownloadedFile(from: location, to: self.destination)
```
ZachNagengast, line 115: I noticed if some files failed to download, it would never attempt to overwrite the existing files, which shouldn't really matter, but this enforces that the blob being downloaded "currently" has all the same files from that version. Future work to be done here related to cache invalidation #39

pcuenca (approved):
Thank you!

PR #65, opened 2024-03-14
CI Tests and SPI Docs

ZachNagengast:
This will run all the tests on PRs. Future work will be to run all the tests on each supported platform. This would be similar to the checks that [swift package index](https://swiftpackageindex.com/huggingface/swift-transformers/builds) is doing, which are now displayed in the readme.

image

Additionally, added the `.spi.yml` needed for swift package index to auto generate docs on their website.

### DIFF
--- a/Sources/TransformersCLI/main.swift
+++ b/Sources/TransformersCLI/main.swift
@@ -53,13 +53,17 @@ struct TransformersCLI: ParsableCommand {
         }
         semaphore.wait()
     }
-    
+
     func compile(at url: URL) throws -> URL {
+        #if os(watchOS)
+        fatalError("Model compilation is not supported on watchOS")
+        #else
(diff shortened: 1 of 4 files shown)


Sources/TransformersCLI/main.swift
```diff
@@ -53,13 +53,17 @@ struct TransformersCLI: ParsableCommand {
         }
         semaphore.wait()
     }
-    
+
     func compile(at url: URL) throws -> URL {
+        #if os(watchOS)
+        fatalError("Model compilation is not supported on watchOS")
+        #else
```
pcuenca, line 60: lol. Maybe we should disable the whole cli on watchOS.

pcuenca (approved):
Fantastic contribution @ZachNagengast, thanks a lot!

Issue #72, opened 2024-03-16
How to use BertTokenizer?


jonathan-goodrx:
what is the best way to use the BertTokenizer? its not a public file so I'm not sure whats the best way to use it

pcuenca:
You are right, @jonathan-goodrx. In general, tokenizers are expected to be used using the `AutoTokenizer.from(pretrained:)` method, but `BertTokenizer` does not yet follow that pattern (some of the pieces are hardcoded instead of relying on `BertPreTokenizer` and `BertNormalizer` configuration components).

For now, I've opened https://github.com/huggingface/swift-transformers/pull/73 to make the class public to unlock your use. Once it's merged, you'd use it like this test shows: https://github.com/huggingface/swift-transformers/blob/3bd02269b7797ade67c15679a575cd5c6f203ce6/Tests/TokenizersTests/BertTokenizerTests.swift#L36. You need to supply the vocabulary mapping.

Would that be helpful for your use case?

In parallel, I'll work in an implementation that follows the same pattern as the other tokenizers in the repo, so it can be used with the `AutoTokenizer.from` methods.

jonathan-goodrx:
that works for me, thank you

jonathan-goodrx:
@pcuenca could you bump the library when you get a chance 🙏

pcuenca:
@jonathan-goodrx Done!

Issue #4, opened 2023-08-08
Tokenizers: additional Normalizers, PreTokenizers, PostProcessors


pcuenca:
So far I've ported the components I needed to support the models I tested, but there are many more in `transformers` and `tokenizers`. For example:

- https://github.com/huggingface/swift-transformers/blob/595c5cfad4107d989f5af7ab06e6d7bbf276c5fc/Sources/Tokenizers/PostProcessor.swift#L27
- https://github.com/huggingface/swift-transformers/blob/595c5cfad4107d989f5af7ab06e6d7bbf276c5fc/Sources/Tokenizers/PreTokenizer.swift#L41

In addition to checking the source code of the [`tokenizers` library](https://github.com/huggingface/tokenizers), I recommend taking a look at the JavaScript implementation by @xenova in [`transformers.js`](https://github.com/xenova/transformers.js/blob/main/src/tokenizers.js) – it's a single file and easy to follow!

jkrukowski:
Added `WhitespacePreTokenizer` here https://github.com/huggingface/swift-transformers/pull/20

011235813:
Adding `BertPreTokenizer` to support BERT-type encoders would be great.

> Tokenizers/PreTokenizer.swift:61: Fatal error: Unsupported PreTokenizer type: BertPreTokenizer

pcuenca:
Thanks @011235813, could you share the tokenizer you are trying to load?

ptsochantaris:
@pcuenca 
```
let tokenizer = try await AutoTokenizer.from(pretrained: "sentence-transformers/all-MiniLM-L6-v2")
```
For example gives an error:
```
Tokenizers/PreTokenizer.swift:66: Fatal error: Unsupported PreTokenizer type: BertPreTokenizer
```

jkrukowski:
> @pcuenca
> 
> ```
> let tokenizer = try await AutoTokenizer.from(pretrained: "sentence-transformers/all-MiniLM-L6-v2")
> ```
> 
> For example gives an error:
> 
> ```
> Tokenizers/PreTokenizer.swift:66: Fatal error: Unsupported PreTokenizer type: BertPreTokenizer
> ```

should be fixed in https://github.com/huggingface/swift-transformers/pull/137

pcuenca:
I think we now have a pretty good coverage of tokenizer building blocks thanks to the community! Closing this issue, feel free to open new ones for missing features.

PR #73, opened 2024-03-18
Make BertTokenizer public

pcuenca:
Temporarily addresses #72

The "right" solution would be to implement the Bert pre-tokenizers, normalizers and Wordpiece model so that Bert tokenizers can be downloaded from configuration files in the Hub like the others.

See also https://github.com/huggingface/swift-transformers/issues/4#issuecomment-1972599789


jonathan-goodrx (approved):
lgtm

PR #80, opened 2024-03-23
add CohereTokenizer support

mzbac:
The CoHere tokenizer is using BPETokenizer. I did some quick tests locally and it seems to work fine with the BPETokenizer.


pcuenca (approved):
Thanks a lot! I need to rework the tokenizer tests so we can easily test all the new tokenizers we add, I'll try to work on that soon.

PR #81, opened 2024-03-23
Support loading tokenizer from local folder #76

### LINKED ISSUE #76: Support loading tokenizer from local folder
Maybe I missed something, but I am currently working on a swift mlx server implementation following the examples at https://github.com/ml-explore/mlx-swift-examples/blob/main/Libraries/LLM/Tokenizer.swift#L7. From my understanding, it seems like the tokenizer configuration has to be loaded via the hub API. I'm wondering if there is a way to load the tokenizer from a local folder instead of fetching it from the Hugging Face repo?

vinu-vanjari:
Resolves #76 

- Added new method in `AutoTokenizer` to load from local model folder. 

- Also added supporting unit tests at `Tests/TokenizersTests/FactoryTests.swift`

- Although we could have simply loaded `tokenizer_config.json` and `tokenizer.json` files directly into configurations, I chose to go via `LanguageModelConfigurationFromHub` because it has it's own magic for choosing appropriate tokenizer class and it also selects fallback tokenizer if needed [refer getter var tokenizerConfig: Config?](https://github.com/huggingface/swift-transformers/blob/4f915610451d29a05948802a140880ff37494dad/Sources/Hub/Hub.swift#L132)

### DIFF
--- a/Sources/Hub/Hub.swift
+++ b/Sources/Hub/Hub.swift
@@ -179,19 +179,9 @@ public class LanguageModelConfigurationFromHub {
     ) async throws -> Configurations {
         let filesToDownload = ["config.json", "tokenizer_config.json", "tokenizer.json"]
         let repo = Hub.Repo(id: modelName)
-        try await hubApi.snapshot(from: repo, matching: filesToDownload)
+        let downloadedModelFolder = try await hubApi.snapshot(from: repo, matching: filesToDownload)
 
-        // Note tokenizerConfig may be nil (does not exist in all models)
-        let modelConfig = try hubApi.configuration(from: "config.json", in: repo)
-        let tokenizerConfig = try? hubApi.configuration(from: "tokenizer_config.json", in: repo)
-        let tokenizerVocab = try hubApi.configuration(from: "tokenizer.json", in: repo)
-        
-        let configs = Configurations(
-            modelConfig: modelConfig,
-            tokenizerConfig: tokenizerConfig,
-            tokenizerData: tokenizerVocab
-        )
-        return configs
+        return try await loadConfig(modelFolder: downloadedModelFolder, hubApi: hubApi)
     }
     
     func loadConfig(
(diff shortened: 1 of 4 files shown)


Sources/Hub/Hub.swift
```diff
@@ -184,6 +193,23 @@ public class LanguageModelConfigurationFromHub {
         )
         return configs
     }
+    
+    func loadConfig(
```
pcuenca, line 197: Can we call this method from `loadConfig(modelName:hubApi:)` instead of duplicating the code? `snapshot` will return the local folder, so we can load from there as you do in the tests :)

Sources/Hub/HubApi.swift
```diff
@@ -94,6 +94,12 @@ public extension HubApi {
     /// `filename` is relative to the download base.
     func configuration(from filename: String, in repo: Repo) throws -> Config {
         let url = localRepoLocation(repo).appending(path: filename)
+        return try configuration(url: url)
+    }
+    
+    /// Assumes the file has already present at local url.
+    /// `url` is complete local file path for given model
```
pcuenca, line 101:
```suggestion
    /// Assumes the file is already present at local url.
    /// `url` is a complete local file path for the given model
```

Sources/Hub/HubApi.swift
```diff
@@ -94,6 +94,12 @@ public extension HubApi {
...
+    func configuration(url: URL) throws -> Config {
```
pcuenca, line 102: Maybe call the parameter `fileURL` to make it clear?

pcuenca (commented):
Very nice, thanks a lot for working on this! 🙌 

This is in very good shape, I only have a couple of suggestions and we can merge.

Maybe we should rename `LanguageModelConfigurationFromHub` in a future PR.

pcuenca (approved):
Looks good! I'll try to trigger a CI run before we merge.

pcuenca:
Something's up with the CI, but the tests pass locally so I'll merge. Thanks a lot @vinu-vanjari!

PR #85, opened 2024-03-29
Add repetition penalty warper

### LINKED ISSUE #84: Is GenerationConfig.repetitionPenalty used during generation?
I am testing the code using the [Core ML version of Llama 2](https://huggingface.co/coreml-projects/Llama-2-7b-chat-coreml).

Setting `GenerationConfig.maxLength` to something larger than the default, e.g., `64`, produces the correct number of output tokens, but tends to repeat tokens towards the end of generation. Adjusting `repetitionPenalty` doesn't seem to have an effect.

Looking into `Generation.swift`, I see the code references `maxLength`, `eosTokenId`, `temperature` and others, but not `repetitionPenalty`.  Does this explain the repetitive output?

shavit:
Closes #84

### DIFF
--- a/Sources/TensorUtils/LogitsWarper/RepetitionPenaltyWarper.swift
+++ b/Sources/TensorUtils/LogitsWarper/RepetitionPenaltyWarper.swift
@@ -6,8 +6,8 @@ import Foundation
 public struct RepetitionPenaltyWarper: LogitsWarper {
     public var penalty: Float
 
-    public init(penalty: Float) {
-        self.penalty = penalty
+    public init(penalty: Double) {
+        self.penalty = Float(penalty)
     }
 
     public func warp(indices: [Int], logits: [Float]) -> (indices: [Int], logits: [Float]) {
--- a/Sources/TensorUtils/LogitsWarper/RepetitionPenaltyWarper.swift
+++ b/Sources/TensorUtils/LogitsWarper/RepetitionPenaltyWarper.swift
@@ -1,0 +1,25 @@
+import Foundation
+
+/// `RepetitionPenaltyWarper` prevents the repetition of previous tokens through a penalty.
+/// This penalty is applied at most once per token.
+/// https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L294
+public struct RepetitionPenaltyWarper: LogitsWarper {
+    public var penalty: Float
+
+    public init(penalty: Float) {
+        self.penalty = penalty
+    }
+
+    public func warp(indices: [Int], logits: [Float]) -> (indices: [Int], logits: [Float]) {
+        var logits = logits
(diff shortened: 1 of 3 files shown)


Sources/TensorUtils/LogitsWarper/RepetitionPenaltyWarper.swift
```diff
@@ -0,0 +1,25 @@
...
+public struct RepetitionPenaltyWarper: LogitsWarper {
+    public var penalty: Float
+
+    public init(penalty: Double) {
+        self.penalty = Float(penalty)
+    }
+
+    public func warp(indices: [Int], logits: [Float]) -> (indices: [Int], logits: [Float]) {
+        var logits = logits
+        for index in indices {
```
pcuenca, line 15: Did you try to find an accelerated version of this? (not critical right now, but good to keep in mind for the future)

Tests/TensorUtilsTests/LogitsWarperTests.swift
```diff
@@ -87,6 +87,36 @@ final class LogitsWarperTests: XCTestCase {
         XCTAssertEqual(result5.logits, [2, 1, 0], accuracy: accuracy)
     }
 
+    func testRepetitionPenaltyWarper() {
```
pcuenca, line 90: Thanks for adding tests! Did you verify results against the reference transformers implementation?

pcuenca (commented):
Thanks a lot! This is a very nice addition to the library 🔥 I'd just like to confirm whether results are the same as in transformers, will run some tests later.

Sources/TensorUtils/LogitsWarper/RepetitionPenaltyWarper.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+/// `RepetitionPenaltyWarper` prevents the repetition of previous tokens through a penalty.
+/// This penalty is applied at most once per token.
+/// https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L294
+public struct RepetitionPenaltyWarper: LogitsWarper {
+    public var penalty: Float
+
+    public init(penalty: Double) {
```
pcuenca, line 9: Maybe let's throw here if `penalty <= 0` (it must be strictly positive)

pcuenca:
Test runners are not working for some reason. I verified locally that the tests pass, so merging now. Thanks @shavit!

Sources/TensorUtils/LogitsWarper/RepetitionPenaltyWarper.swift
shavit: I wanted to use gather scatter, but didn't find scatter method in vDSP.

PR #86, opened 2024-03-30
[WIP] Create embedding module

### LINKED ISSUE #22: Support for embedding models (BGE, GTE etc)
Embeddings models are very useful, and can easily be run on device in terms of hardware specs.

It would be awesome if swift-transformers would work with them.

shavit:
Closes #22

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,7 +7,7 @@ let package = Package(
     name: "swift-transformers",
     platforms: [.iOS(.v16), .macOS(.v13)],
     products: [
-        .library(name: "Transformers", targets: ["Tokenizers", "Generation", "Models"]),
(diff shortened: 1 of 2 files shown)


shavit:
Right, I’ve got the BERT embeddings implemented, but this pull request relies on another one for reading safe tensors. That said, having the BERT tokenizer implemented should be sufficient, and I've already done that in a different branch [<branch URL>](https://github.com/shavit/swift-transformers/tree/4-pretokenizer-bert-unsupported), and merged both into another project.

mattt:
Hi @shavit. Thank you for laying the groundwork for embeddings here. I'm going to close this out to keep the queue tidy since it's been a WIP for a while.

As discussed in https://github.com/huggingface/swift-transformers/issues/22, there are now good options out there for generating embeddings, like https://github.com/jkrukowski/swift-embeddings.

Thanks again for your work on this!

PR #89, opened 2024-04-11
Supporting more BERT-like models

ashvardanian:
Hi HF team!

I am extending our UForm repository of multimodal models to support Swift and mobile deployments, and along that way I've noticed that several classes for a broad range of BERT-like models are not yet supported by `swift-transformers`. So I've added a `WordPieceDecoder` class and aliases for `BertPreTokenizer` and `BertProcessing`.

Moreover, are you are well aware `config.json` and `tokenizer.json` come in all shapes and sizes. So I've added fallback mechanisms to handle different tuple order in vocabulary listings.

The [current `main-dev` branch of UForm](https://github.com/unum-cloud/uform/commit/00bd84c59995c7b3daa0b4fa1597f77608806fdb) is already using this functionality from my fork. I am looking into integrating more `Hub` functionality next. Please let me know what you think about this PR 🤗

### DIFF
--- a/Sources/Hub/Hub.swift
+++ b/Sources/Hub/Hub.swift
@@ -100,7 +100,14 @@ public struct Config {
     }
     
     /// Tuple of token identifier and string value
-    public var tokenValue: (UInt, String)? { value as? (UInt, String) }
+    public var tokenValue: (UInt, String)? {
+        if let tuple = value as? (UInt, String) {
+            return tuple
+        } else if let reversedTuple = value as? (String, UInt) {
+            return (reversedTuple.1, reversedTuple.0)
+        }
+        return nil
+    }
 }
 
 public class LanguageModelConfigurationFromHub {
--- a/Sources/Hub/Hub.swift
+++ b/Sources/Hub/Hub.swift
@@ -101,11 +101,16 @@ public struct Config {
     
     /// Tuple of token identifier and string value
     public var tokenValue: (UInt, String)? {
-        if let tuple = value as? (UInt, String) {
-            return tuple
-        } else if let reversedTuple = value as? (String, UInt) {
-            return (reversedTuple.1, reversedTuple.0)
+        guard let array = value as? [Any], array.count == 2 else {
+            return nil
         }
+
+        if let first = array[0] as? String, let second = array[1] as? Int64 {
+            return (UInt(second), first)
+        } else if let first = array[0] as? Int64, let second = array[1] as? String {
+            return (UInt(first), second)
+        }
+
         return nil
     }
 }
--- a/Sources/Tokenizers/PostProcessor.swift
+++ b/Sources/Tokenizers/PostProcessor.swift
@@ -22,20 +22,30 @@ extension PostProcessor {
 }
 
 enum PostProcessorType: String {
-    case TemplateProcessing
+    case Template
     case ByteLevel
-    case RobertaProcessing
+    case Bert
+    case Roberta
+
+    static let BertProcessing = "Bert"
+    static let RobertaProcessing = "Roberta"
+    static let TemplateProcessing = "Template"
 }
 
 struct PostProcessorFactory {
     static func fromConfig(config: Config?) -> PostProcessor? {
         guard let config = config else { return nil }
-        guard let typeName = config.type?.stringValue else { return nil }
+        guard var typeName = config.type?.stringValue else { return nil }
+        if typeName.hasSuffix("Processing") {
+            typeName = String(typeName.dropLast("Processing".count))
+        }
+
         let type = PostProcessorType(rawValue: typeName)
         switch type {
-        case .TemplateProcessing: return TemplateProcessing(config: config)
+        case .Template          : return TemplateProcessing(config: config)
         case .ByteLevel         : return ByteLevelPostProcessor(config: config)
-        case .RobertaProcessing : return RobertaProcessing(config: config)
+        case .Bert              : return BertProcessing(config: config)
+        case .Roberta           : return RobertaProcessing(config: config)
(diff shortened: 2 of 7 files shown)


ashvardanian:
Another open problem that I've recently discovered is the way strings are compared in Swift. By default, the language uses UTF8-aware normalization techniques when comparing strings. This is great for some applications, but horrible for tokenization, especially with multilingual models. I've solved that by introducing a `LiteralString` wrapper for `String`, that uses the literal comparators:

```swift
    struct LiteralString: Hashable {
        let value: String

        static func ==(lhs: LiteralString, rhs: LiteralString) -> Bool {
            return lhs.value.compare(rhs.value, options: .literal) == .orderedSame
        }

        func hash(into hasher: inout Hasher) {
            hasher.combine(value)
        }
    }
```

I believe it should be applicable in other places as well. Let me know what you think, @pcuenca 🤗

Sources/Tokenizers/PostProcessor.swift
```diff
@@ -22,20 +22,30 @@ extension PostProcessor {
...
-    case TemplateProcessing
+    case Template
     case ByteLevel
-    case RobertaProcessing
+    case Bert
+    case Roberta
+
+    static let BertProcessing = "Bert"
+    static let RobertaProcessing = "Roberta"
+    static let TemplateProcessing = "Template"
```
pcuenca, line 32: I'm not sure this is worth doing, I find the aliases and the suffix removal distracting for little benefit. My original approach was to simply use the same names that appear in the json, so someone reading both could easily match.

Sources/Tokenizers/UnigramTokenizer.swift
```diff
@@ -50,11 +65,14 @@ class UnigramTokenizer: PreTrainedTokenizerModel {
         self.unknownTokenId = unknownTokenId
         self.unknownPiece = SentencePieceToken(token: vocab[unknownTokenId].token, score: minScore - 10)
         
-        tokensToIds = Dictionary(uniqueKeysWithValues: vocab.map { $0.token }.enumerated().map { ($1, $0) })
-        bosTokenId = tokensToIds[bosToken!]      // May be nil
+        // Using `Dictionary(uniqueKeysWithValues:)` is the default approach for constructing the mapping.
+        // It, however, will use the normal `compare` function of strings and will result in collisions for different 
+        // UTF8 strings. Instead, we should use the `a.compare(b, options: .literal) == .orderedSame`.
+        tokensToIds = Dictionary(uniqueKeysWithValues: vocab.map { $0.token }.enumerated().map { (LiteralString(value: $1), $0) })
+        bosTokenId = tokensToIds[LiteralString(value: bosToken!)]      // May be nil
```
pcuenca, line 72: maybe worth considering a `String` extension that returns a `LiteralString`, for readability? (Just an idea)

ashvardanian: Maybe 🤔

Sources/Hub/Hub.swift
```diff
@@ -100,7 +100,19 @@ public struct Config {
...
+        }
+
+        if let first = array[0] as? String, let second = array[1] as? Int64 {
+            return (UInt(second), first)
+        } else if let first = array[0] as? Int64, let second = array[1] as? String {
+            return (UInt(first), second)
+        }
+
+        return nil
+    }
```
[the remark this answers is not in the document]
ashvardanian: Sure, I was testing everything on our models, and we generally don't invent new config structures and reuse existing variants.

image

We are preparing a [release of the re-packaged encoders](https://huggingface.co/collections/unum-cloud/uform-3-encoders-662344ee44bc96a4aa5937cd). They work fine with Python and JavaScript Transformers libraries, but required all of these changes to work in Swift 🤗

Sources/Tokenizers/PostProcessor.swift
ashvardanian: It's very hard to use JSON as the ground truth, as they cone in all shapes and sizes. 

Shorter name was needed for our models to work, but I've added the static variables for backward-compatibility, to avoid breaking the library for other users.

Sources/Hub/Hub.swift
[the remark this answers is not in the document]
pcuenca: Yes, I understand! It was for documentation purposes only (maybe as a comment if appropriate) :) But it's ok!

Sources/Tokenizers/PostProcessor.swift
pcuenca: I'd say the json names used in transformers should be quite stable right now. Why do your models require shorter names? (just curious)

ashvardanian: I guess no model "requires" specific names, and JSONs can always be changed, but it generally results in a snowballing set of changes that have to be applied on every platform...

In UForm I have identical tests that run for the same models on the same data across all 3 languages across ONNX, PyTorch, and CoreML. You can [check them here](https://github.com/ashvardanian/uform/blob/main/CONTRIBUTING.md).

If a certain behavior is standard in the more popular ports of the library (Python and JS), I assume Hugging Face may want to provide the same behavior here to encourage adoption. A lot of people would probably appreciate the portability 🤗

ConfuseIous:
Any chance of this being merged in soon? I'm trying to use a BERT model and this PR would be a huge help :)

eemilk:
I was trying out these changes since have DistilBERT model I wanted to test, but getting thrown `unsupportedTokenizer("Distilbert")` on `try await AutoTokenizer.from(pretrained:..`

Is there a chance support for DistilBERT is also added?

pcuenca:
Superseded by #137

PR #90, opened 2024-04-11
Automatic Formatting with `apple/swift-format`

ashvardanian:
When working with the repository, I am not always sure about how to style code, as Swift gives a lot of flexibility on where to put `public` clauses, how to indent the code and so on. Moreover, Xcode isn't friendly to "Format on Save". Luckily, there is a `swift-format` utility from Apple, that works very similar to `clang-format`.

We extensively use such systems in our codebases. They help keep the code sane when the number of contributors rises. If interested, you may want to add it to your Git hooks.

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -1,0 +1,13 @@
+{
+    "version": 1,
+    "lineLength": 120,
+    "indentation": {
+        "spaces": 4
+    },
+    "maximumBlankLines": 1,
+    "respectsExistingLineBreaks": true,
+    "lineBreakBeforeControlFlowKeywords": true,
+    "lineBreakBeforeEachArgument": true,
+    "multiElementCollectionTrailingCommas": true,
+    "spacesAroundRangeFormationOperators": true
+}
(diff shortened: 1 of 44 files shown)


maiqingqiang:
Nice! I've been waiting.

pcuenca:
Thank you, this was surely needed! I'll review in depth soon. Would you say that the rules you chose are more or less standard across Swift codebases, or are there other alternatives? (I personally don't like to break up all function parameters in multiple lines, but willing to sacrifice for consistency).

ashvardanian:
Not sure the standard exists in Swift. I'm now enabling these rules in the libraries I maintain - UForm, USearch, StringZilla. Maybe with enough popular libraries agreeing on the format, we can propose that missing standard 🤷‍♂️

maiqingqiang:
Xcode 16 has a built-in apple/swift-format

https://www.reddit.com/r/iOSProgramming/comments/1df0ooe/xcode_16_now_has_a_builtin_formatter/

PR #91, opened 2024-04-20
Read model weights 

shavit:
Read model weights from safetensors.

Related https://github.com/huggingface/swift-transformers/pull/86


shavit:
Weights are now accessible independently of configurations for flexibility and without MLModel or conversion.

pcuenca:
That's very cool @shavit! May I ask what use case you have in mind for this?

shavit:
Yes
  * In the first commit I added weights property to `Configurations`, but since `LanguageModel` requires a model, I separated the weights initialization.
   * The weights can go into the `Embedding` module instead, since the language models in this module are tied to CoreML, and don't need this at the moment.
   * The `Safetensor` and others can save updated weights into a file.

The plan is to use models directly from local files or the hub, and maybe support more types like mlx/numpy.

Tests/TensorUtilsTests/WeightsTests.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    let downloadDestination: URL = {
+        FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!.appending(component: "huggingface-tests")
+    }()
+
+    var hubApi: HubApi { HubApi(downloadBase: downloadDestination) }
+
+    func testLoadWeightsFromFileURL() async throws {
+        let repo = "google/bert_uncased_L-2_H-128_A-2"
+        let modelDir = try await hubApi.snapshot(from: repo)
```
pcuenca, line 15: I'd suggest you use [matching globs](https://github.com/huggingface/swift-transformers/blob/37e234e1ee4f2ef9c21c3d5cd592b5ec63d879af/Tests/HubTests/HubApiTests.swift#L188) to only download the two files we need and avoid additional downloads. There's a [version](https://github.com/huggingface/swift-transformers/blob/37e234e1ee4f2ef9c21c3d5cd592b5ec63d879af/Sources/Hub/HubApi.swift#L204) that supports multiple patterns.

pcuenca (approved):
Very cool work, thanks a lot for contributing! I only have one suggestion regarding one of the tests.

pcuenca (approved):
Ready to merge, in my opinion! Are you planning any more tweaks?

shavit:
No, that's it.

pcuenca:
Thanks a lot @shavit!

PR #93, opened 2024-04-26
Added Tokens

### LINKED ISSUE #92: Ensure added tokens are supported

pcuenca:
Fixes #92

### DIFF
--- a/Sources/Tokenizers/PreTokenizer.swift
+++ b/Sources/Tokenizers/PreTokenizer.swift
@@ -248,7 +248,7 @@ extension StringSplitPattern {
     }
 }
 
-extension String {
+public extension String {
     func ranges(of string: String, options: CompareOptions = .regularExpression) -> [Range<Index>] {
         var result: [Range<Index>] = []
         var start = startIndex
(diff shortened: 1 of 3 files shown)


Sources/Tokenizers/PreTokenizer.swift
```diff
@@ -9,27 +9,26 @@ import Foundation
...
 
 public protocol PreTokenizer {
-    func preTokenize(text: String) -> [String]
-    func preTokenize(texts: [String]) -> [String]
-    func callAsFunction(texts: [String]) -> [String]
-    func callAsFunction(text: String) -> [String]
+    func preTokenize(text: String, firstSection: Bool) -> [String]
+    func preTokenize(texts: [String], firstSection: Bool) -> [String]
+    func callAsFunction(texts: [String], firstSection: Bool) -> [String]
+    func callAsFunction(text: String, firstSection: Bool) -> [String]
```
pcuenca, line 15: Unfortunately I had to change this public protocol so that `MetaspacePreTokenizer` can conditionally work depending on whether we are handling the first split. This is necessary because the input text is split by the added tokens first, and then all the chunks go through the tokenization pipeline. Do you think this will break something in your codebases @ZachNagengast @awni?

Sources/Tokenizers/PreTokenizer.swift
```diff
@@ -277,6 +276,42 @@ extension String {
         return result
     }
 
+    /// This version supports capture groups, wheres the one above doesn't
+    func split(by captureRegex: NSRegularExpression) -> [String] {
```
pcuenca, line 280: This was a bit annoying, it's just another version of the previous `split` helper we wrote that works with capture groups.

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -241,9 +262,32 @@ public class PreTrainedTokenizer: Tokenizer {
 
 public struct AutoTokenizer {}
 
+struct PreTrainedTokenizerClasses {
+    /// Class overrides for custom behaviour
+    /// Not to be confused with the TokenizerModel classes defined in TokenizerModel
+    static let tokenizerClasses: [String : PreTrainedTokenizer.Type] = [
+        "LlamaTokenizer": LlamaPreTrainedTokenizer.self
```
pcuenca, line 269: There can be tokenizer subclasses now (not just tokenizer models)

Sources/Tokenizers/PreTokenizer.swift
awni: @davidkoski wdyt?

awni:
Thanks for the quick fix @pcuenca 🚀 🤗

Tests/TokenizersTests/AddedTokensTests.swift
```diff
@@ -0,0 +1,67 @@
...
+//
+//  Created by Pedro Cuenca on 20240426.
+//
+
+import XCTest
+import Tokenizers
+import Hub
+
+class AddedTokensTests: XCTestCase {
+    func testPhiAddedEnd() async throws {
```
pcuenca, line 13: This test now passes, and it produces the same results as Python.

Sources/Tokenizers/PreTokenizer.swift
pcuenca: A default value is provided in a protocol extension, but the protocol definition itself cannot have default values, unfortunately.

pcuenca:
The CI workers are down, but all tests pass locally. I'll wait a few hours to merge unless you think it's urgent.

Also cc @DanThePutzer, @MatthewWaller in case they want to test or review.

Sources/Tokenizers/PreTokenizer.swift
davidkoski: looks fine for mlx-swift.  I wonder if this should be an enum instead of a bool?  I don't know if there are other "contexts" that might be used...

pcuenca: Good call @davidkoski, I did check the transformers.js port and didn't see any other context options, but better to play it safe.

PR #94, opened 2024-04-28
Allow unigram tokenizer to handle NSNumber score values 

ZachNagengast:
For whatever reason, certain config files for unigram tokenizers end up reading scores as NSNumber instead of float, and throw malformedVocab errors.

`Expected a Float but found __NSCFNumber`

This doesn't convert directly to Float, so I'm adding one fallback here to attempt to convert it to float before throwing.


pcuenca:
Interesting! Maybe because of precision issues?

pcuenca (approved):
Looks good to me, I'll run the tests locally.

PR #101, opened 2024-06-16
Strict concurrency in Swift 6

DePasqualeOrg:
This is a starting point for strict concurrency in Swift 6. There are still some errors and warnings, which I'm not sure how to fix, so perhaps someone more knowledgeable than me can take it from here.


DePasqualeOrg:
https://github.com/ml-explore/mlx-swift-examples/pull/90

Issue #105, opened 2024-07-15
Update to swift-argument-parser 1.4


awni:
@pcuenca is it possilbe to upgrade the dependency?

https://github.com/huggingface/swift-transformers/blob/main/Package.swift#L15

Would be useful for us to have it on >= 1.4 to pick up concurrency changes.

PR #106, opened 2024-07-16
Update swift-argument-parser to >= 1.4.0

pcuenca:
Addresses #105

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,7 +12,7 @@ let package = Package(
         .executable(name: "hub-cli", targets: ["HubCLI"]),
     ],
     dependencies: [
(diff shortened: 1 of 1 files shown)


pcuenca:
cc @awni 

Sorry, I'm travelling and a bit slow to respond. Happy to merge and create a release, assuming it has no backwards compatibility issues.

awni:
Thanks @pcuenca for the quick fix. If you are worried about back-compatability, I think `from: "1.3.0"` will also work. CC @davidkoski.

Would be great if you could merge and release with this patch!

ZachNagengast (approved):

PR #119, opened 2024-09-04
Fix Jinja Version

maiqingqiang:
I'm very sorry, I didn't tag Jinja, which caused the inability to properly depend on swift-transformers 0.1.10.

```
Failed to resolve dependencies Dependencies could not be resolved because root depends on 'swift-transformers' 0.1.10..<1.0.0.
'swift-transformers' >= 0.1.10 cannot be used because package 'swift-transformers' is required using a stable-version but 'swift-transformers' depends on an unstable-version package 'jinja' and no versions of 'swift-transformers' match the requirement 0.1.11..<1.0.0.
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/maiqingqiang/Jinja", branch: "main")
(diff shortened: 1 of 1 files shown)


pcuenca (approved):

pcuenca:
Pushed tag 0.1.11 as well

maiqingqiang:
> Pushed tag 0.1.11 as well

Thank you very much. 

image

PR #124, opened 2024-09-19
Upgrade Jinja 1.0.1 - Fix Qwen2.5 Chat Template error

maiqingqiang:
Fix Qwen2.5 Chat Template error: evaluateUnaryExpression Unknown argument type: UndefinedValue https://github.com/maiqingqiang/Jinja/commit/4a6eb1a94ef54cf6a57c6325a13d73295626dd8c

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.1")
+        .package(url: "https://github.com/maiqingqiang/Jinja", "1.0.1"..<"1.1.0")
     ],
     targets: [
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.0")
+        .package(url: "https://github.com/maiqingqiang/Jinja", "1.0.1"..<"1.1.0")
```
pcuenca, line 16: Hmm, why is this necessary? If you run `Update to Latest Package Versions` in Xcode it retrieves `1.0.1`, as expected.

maiqingqiang: Haha, I just found out that trick too! 😂
Thanks for the heads-up!

PR #125, opened 2024-09-20
Replace usage of Bundle.module to build outside SPM context

a2they:
Replacing `Bundle.module` with `Bundle(for: self)` to accommodate building transformers outside SPM context.

### DIFF
--- a/Sources/Hub/Hub.swift
+++ b/Sources/Hub/Hub.swift
@@ -202,7 +202,8 @@ public class LanguageModelConfigurationFromHub {
     }
 
     static func fallbackTokenizerConfig(for modelType: String) -> Config? {
-        guard let url = Bundle.module.url(forResource: "\(modelType)_tokenizer_config", withExtension: "json") else { return nil }
+        let bundle = Bundle(for: self)
+        guard let url = bundle.url(forResource: "\(modelType)_tokenizer_config", withExtension: "json") else { return nil }
         do {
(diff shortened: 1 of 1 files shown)


pcuenca (approved):
Very cool, thanks for showing how to do it! 🔥 

Apologies for the delay in reviewing, it was a 🦙 week.

I'll check what's happening with the CI.

Sources/Hub/Hub.swift
```diff
@@ -202,7 +202,8 @@ public class LanguageModelConfigurationFromHub {
     }
 
     static func fallbackTokenizerConfig(for modelType: String) -> Config? {
-        guard let url = Bundle.module.url(forResource: "\(modelType)_tokenizer_config", withExtension: "json") else { return nil }
+        let bundle = Bundle(for: self)
```
pcuenca, line 205: We could fix the tests doing something like:

```suggestion
        let bundle = NSClassFromString("XCTest") != nil ? Bundle.module : Bundle(for: self)
```

But upon additional testing I found that building a CLI that uses `Tokenizers` would not find the fallback configuration either.

Do you have any thoughts on that? Some additional ideas:
- Iterate through the bundles and select the right one.
- Create a bundle (or framework?) solely dedicated to hosting the resources.
- Download the fallback configs from a well-known location in the Hub.

pcuenca:
This turned out trickier than expected, closing for now.

PR #129, opened 2024-09-25
Add Sequence and Bert processors

DePasqualeOrg:
Ported from the Rust implementations here: https://github.com/huggingface/tokenizers/tree/main/tokenizers/src/processors

I've tested this with Llama 3.2 1B and 3B, and it works. I haven't been able to test the Bert processor. Also, this is coming from Claude 3.5 Sonnet, and I don't understand all the details, so please check this carefully.

### DIFF
--- a/Sources/Tokenizers/PostProcessor.swift
+++ b/Sources/Tokenizers/PostProcessor.swift
@@ -25,6 +25,8 @@ enum PostProcessorType: String {
     case TemplateProcessing
     case ByteLevel
     case RobertaProcessing
+    case BertProcessing
+    case Sequence
 }
 
 struct PostProcessorFactory {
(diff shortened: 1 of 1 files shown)


pcuenca:
Thanks a lot @DePasqualeOrg 🙌 Taking a look.

Sources/Tokenizers/PostProcessor.swift
```diff
@@ -139,3 +143,51 @@ class RobertaProcessing: PostProcessor {
...
+
+    func postProcess(tokens: [String], tokensPair: [String]?, addSpecialTokens: Bool = true) -> [String] {
+        var currentTokens = tokens
+        var currentTokensPair = tokensPair
+
+        for processor in processors {
+            let processed = processor.postProcess(tokens: currentTokens, tokensPair: currentTokensPair, addSpecialTokens: addSpecialTokens)
+            currentTokens = processed
+            currentTokensPair = nil  // After the first processor, we no longer have a separate pair
+        }
```
pcuenca, line 189: The implementation in [transformers.js is slightly different](https://github.com/xenova/transformers.js/commit/db9250b3e814dc43b1cb35d9c2cd0feff30e3932#diff-323fc64a0a2174a18ce315f2bc15d7ef8ac9c34e68c394f9eba64516e61aeb41R1773) cc @xenova

pcuenca (approved):
I ran some additional tests locally (I can push to a followup PR) and it works fine, so I'm approving the PR 🚀 

However, I saw a couple issues we should investigate.
- The implementation of the sequence post-processor is [slightly different in transformers.js](https://github.com/xenova/transformers.js/commit/db9250b3e814dc43b1cb35d9c2cd0feff30e3932#diff-323fc64a0a2174a18ce315f2bc15d7ef8ac9c34e68c394f9eba64516e61aeb41R1773), as noted in a comment. Not sure what the implications are yet.
- Tokenizers >= 0.20.0 serializes BPE merges as a `[[String]]` instead of a `[String]`, which resolves the ambiguity for merges containing spaces. However, this format is not supported by `swift-transformers`. For example, [this tokenizer](https://huggingface.co/pcuenq/Llama-3.2-1B-Instruct-tokenizer) will currently fail. I will address this separately as well, as I expect this serialization to become more frequent going forward.

Thanks again for the super fast contribution @DePasqualeOrg! 🔥

PR #131, opened 2024-09-26
Support new merges serialization

pcuenca:
Introduced in tokenizers 0.20.0. Tokenizers saved with it will create a `merges` property where each merge is an array of two items, instead of a string with a separator.


pcuenca:
This is confirmed to be the new export format, merging.

PR #134, opened 2024-10-01
Support multiple chat templates per model

DePasqualeOrg:
Some models, such as [Mistral 7B Instruct 0.3](https://huggingface.co/mlx-community/Mistral-7B-Instruct-v0.3-4bit/blob/main/tokenizer_config.json),  specify an array of chat templates in `tokenizer_config.json`. This change allows for parsing chat templates in this format.

### DIFF
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -125,7 +125,8 @@ public protocol Tokenizer {
         chatTemplate: String?,
         addGenerationPrompt: Bool,
         truncation: Bool,
-        maxLength: Int?
+        maxLength: Int?,
+        tools: [[String: Any]]?
     ) throws -> [Int]
 }
 
@@ -323,12 +324,35 @@ public class PreTrainedTokenizer: Tokenizer {
     
     public func applyChatTemplate(
         messages: [[String: String]],
-        chatTemplate: String?,
+        chatTemplate: String? = nil,
         addGenerationPrompt: Bool = false,
         truncation: Bool = false,
-        maxLength: Int?
+        maxLength: Int? = nil,
+        tools: [[String: Any]]? = nil
     ) throws -> [Int] {
-        let template = try Template(chatTemplate ?? tokenizerConfig.chatTemplate?.stringValue ?? defaultChatTemplate)
+        var chatTemplateFromConfig: String?
+        if let chatTemplateValue = tokenizerConfig.chatTemplate {
+            if let chatTemplateStringValue = chatTemplateValue.stringValue {
+                chatTemplateFromConfig = chatTemplateStringValue
+            } else if let chatTemplateArrayValue = chatTemplateValue.arrayValue {
+                // If a list of chat templates is specified, convert them to a dict
+                let templateDict = Dictionary<String, String>(uniqueKeysWithValues: chatTemplateArrayValue.compactMap { template in
+                    guard let name = template[dynamicMember: "name"]?.stringValue,
+                          let templateString = template[dynamicMember: "template"]?.stringValue else {
+                        return nil
+                    }
+                    return (name, templateString)
+                })
+                // Choose the appropriate template
+                if let tools = tools, !tools.isEmpty, let toolUseTemplate = templateDict["tool_use"] {
+                    chatTemplateFromConfig = toolUseTemplate
+                } else {
+                    chatTemplateFromConfig = templateDict["default"]
+                }
+            }
+        }
+
+        let template = try Template(chatTemplate ?? chatTemplateFromConfig ?? defaultChatTemplate)
         var context: [String: Any] = [
             "messages": messages,
             "add_generation_prompt": addGenerationPrompt
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -94,6 +94,13 @@ struct TokenizerModel {
     }
 }
 
+public enum ChatTemplateArgument {
+    /// A Jinja template to use for the conversion. Normally it is not necessary to provide a template, since it will be read from the tokenizer config file.
+    case literal(String)
+    /// For models whose tokenizer config file includes multiple chat templates, the template can be specified by name. Normally this is not necessary.
+    case name(String)
+}
+
 public protocol Tokenizer {
     func tokenize(text: String) -> [String]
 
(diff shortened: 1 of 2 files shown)


DePasqualeOrg:
I've tried to bring this more in line with the [Python implementation](https://github.com/huggingface/transformers/blob/e5d14f39ad82475f238cad41279a5d61ac5db287/src/transformers/tokenization_utils_base.py#L1707). I'm unsure about whether `tools` should be included in the `context` dictionary.

pcuenca:
> I'm unsure about whether tools should be included in the context dictionary.

What do you mean? As I understand it, you are adding support to the Jinja Swift package [here](https://github.com/maiqingqiang/Jinja/pull/4), right?

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -323,12 +324,35 @@ public class PreTrainedTokenizer: Tokenizer {
...
+                } else {
+                    chatTemplateFromConfig = templateDict["default"]
+                }
+            }
+        }
+
+        let template = try Template(chatTemplate ?? chatTemplateFromConfig ?? defaultChatTemplate)
         var context: [String: Any] = [
             "messages": messages,
             "add_generation_prompt": addGenerationPrompt
```
DePasqualeOrg, line 358: Should `tools` be included here?

DePasqualeOrg, line 358: Should `tools` be included here?

DePasqualeOrg:
> > I'm unsure about whether tools should be included in the context dictionary.
> 
> What do you mean? As I understand it, you are adding support to the Jinja Swift package [here](https://github.com/maiqingqiang/Jinja/pull/4), right?

I think that PR won't include the tools functionality, for the sake of simplicity. Later on I may try to add it.

pcuenca:
> I think that PR won't include the tools functionality

In that case, maybe we should get started with just a simple way to select the desired template (by name) and add the tool stuff later. Would that work for the purposes of https://github.com/ml-explore/mlx-swift-examples/pull/135?

DePasqualeOrg:
I think this now mirrors the Python implementation more closely. It throws an error if no template is selected, which I think is better than silently falling back to a default template that might not be compatible with the model, which previously caused some confusion for me when I was debugging the problems with Mistral 7B 0.3.

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -323,12 +324,35 @@ public class PreTrainedTokenizer: Tokenizer {
...
+            } else if let chatTemplateArrayValue = chatTemplateValue.arrayValue {
+                // If a list of chat templates is specified, convert them to a dict
+                let templateDict = Dictionary<String, String>(uniqueKeysWithValues: chatTemplateArrayValue.compactMap { template in
+                    guard let name = template[dynamicMember: "name"]?.stringValue,
+                          let templateString = template[dynamicMember: "template"]?.stringValue else {
+                        return nil
+                    }
+                    return (name, templateString)
+                })
+                // Choose the appropriate template
```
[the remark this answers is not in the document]
DePasqualeOrg: Done.

DePasqualeOrg:
I've added some tests with a note to add tests for tool use templates once these are implemented. Do you think this can be merged without implementing tool use in Jinja? It would be an improvement on the current situation, in which Mistral 7B doesn't work properly.

pcuenca:
> I've added some tests with a note to add tests for tool use templates once these are implemented.

Thank you!

> Do you think this can be merged without implementing tool use in Jinja? It would be an improvement on the current situation, in which Mistral 7B doesn't work properly.

Yes, let's merge this and we can improve later. I have a couple of final comments / questions and that's it.

DePasqualeOrg:
I added an extra argument `templateName` to make things more explicit. Otherwise it becomes difficult to cover edge cases. Do you think it's okay to diverge slightly from the Python implementation here?

DePasqualeOrg:
We can maintain consistency with the Python API while being more explicit using an enum for the `chatTemplate` argument, with options for `literal` and `name`.

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -323,12 +323,55 @@ public class PreTrainedTokenizer: Tokenizer {
     
     public func applyChatTemplate(
         messages: [[String: String]],
-        chatTemplate: String?,
+        /// A Jinja template or the name of a template to use for this conversion.
+        /// It is usually not necessary to pass anything to this argument,
+        /// as the model's template will be used by default.
+        chatTemplate: String? = nil,
         addGenerationPrompt: Bool = false,
```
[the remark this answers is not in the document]
pcuenca: I think it should be `true` for the overloads as I expect that to be the most common use of the API, but I'm not fully sure.

pcuenca:
> using an enum

That's a great idea!

In any case, it is ok to deviate a bit from the Python API if it's useful for the project. I see it more as a reference. Also, the more we use the same criteria, the less friction people that come from Python will experience. But I think that some developers don't come from that audience.

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -94,6 +94,13 @@ struct TokenizerModel {
     }
 }
 
+public enum ChatTemplateArgument {
+    /// A Jinja template to use for the conversion. Normally it is not necessary to provide a template, since it will be read from the tokenizer config file.
```
pcuenca, line 98:
```suggestion
    /// A Jinja template to use for the conversation. Normally it is not necessary to provide a template, since it will be read from the tokenizer config file.
```

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -117,17 +124,15 @@ public protocol Tokenizer {
...
     var unknownToken: String? { get }
     var unknownTokenId: Int? { get }
-    
-    func applyChatTemplate(messages: [[String: String]]) throws -> [Int]
 
-    func applyChatTemplate(messages: [[String: String]], chatTemplate: String) throws -> [Int]
+    func applyChatTemplate(messages: [[String: String]]) throws -> [Int]
 
-    func applyChatTemplate(messages: [[String: String]], chatTemplateName: String) throws -> [Int]
+    func applyChatTemplate(messages: [[String: String]], chatTemplate: ChatTemplateArgument) throws -> [Int]
```
pcuenca, line 130: I like this solution. But it's a breaking change, do you know if the current method is used by mlx or others?

Perhaps we could add an overload `applyChatTemplate(messages: [[String: String]], chatTemplateName: String)` that just builds a `ChatTemplateArgument` (`literal`) and forwards the call.

pcuenca (commented):
Looking great, thanks a lot for your patience!

Sources/Tokenizers/Tokenizer.swift
DePasqualeOrg: I searched the mlx-swift and mlx-swift-examples repos and didn't find any instances of `applyChatTemplate` there. I have a [draft PR](https://github.dev/ml-explore/mlx-swift-examples) to use the chat templates once this PR and the Jinja PR land.

I think you mean the additional overload should have the argument `chatTemplate: String`. Sure, I can add that.

pcuenca (approved):
Thanks a lot for iterating here @DePasqualeOrg 🙌 

Great contribution, merging this now!

DePasqualeOrg:
Thanks for your help! Do you want to make a new release so that this can get picked up in mlx-swift-examples for https://github.com/ml-explore/mlx-swift-examples/pull/135?

pcuenca:
Just pushed https://github.com/huggingface/swift-transformers/releases/tag/0.1.13

PR #135, opened 2024-10-04
Make `progressHandler` in `snapshot` `@Sendable`

DePasqualeOrg:
This resolves concurrency warnings with complete concurrency checking.


DePasqualeOrg:
Actually, I can also resolve the error by marking the closure `@Sendable` when I call `snapshot`, so maybe this is not needed.

PR #146, opened 2024-12-24
Add Qwen2Tokenizer

finnvoor:
Tried to port the python tokenizer but didn't really know how so I used the BPETokenizer like most of the other models and it seems to work 🤷


finnvoor:
Not quite sure how to generate the test resources / if a test is necessary

pcuenca (approved):
Hi @finnvoor, thanks a lot! I tested locally applying https://huggingface.co/Qwen/Qwen2.5-1.5B-Instruct to the test cases we usually run, and it worked.

We could add tests in a new PR, but I think it's not necessary at this point because:
- This tokenizer is built on the same components as many others.
- We should rework the testing infrastructure to make it easier for contributors to add more tests without bloating the test resources.

Thank you!

PR #147, opened 2024-12-24
Metadata test: use "resolve" URL

pcuenca:


pcuenca:
The previous URL was retrieving metadata for the _HTML page_ that is rendered in the website, not the file itself.

Perhaps we should define another API to avoid confusion, cc @ardaatahan

PR #148, opened 2024-12-25
Add skipSpecialTokens option to Tokenizer.decode

finnvoor:
Merry Christmas! 🎅

This PR adds a `skipSpecialTokens` option to `Tokenizer.decode`. Default arguments can't be used since Tokenizer is a protocol, so I added another function, leaving `func decode(tokens:)` as a convenience / maintain source compatibility.


pcuenca (approved):
Looks great to me, thank you @finnvoor 🙌, and merry Christmas!

PR #150, opened 2024-12-31
Auto-format with swift-format

DePasqualeOrg:
I copied the swift-format and pre-commit configs that were recently [added](https://github.com/johnmai-dev/Jinja/pull/10) to Jinja. I also added a GitHub action to check formatting. This should minimize distracting diffs due to formatting. Feel free to change the formatting rules if you have a strong preference about them.

For future PRs, formatting can be applied with `pre-commit run --all-files` before committing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+- repo: https://github.com/slessans/pre-commit-swift-format
+  rev: "fd627de92bdf84a75c924ed95691336d14e94cf1"
+  hooks:
+    - id: swift-format
+      args: ["--configuration", ".swift-format"]
--- a/.swift-format
+++ b/.swift-format
@@ -1,0 +1,9 @@
+{
+  "version": 1,
+  "indentation": {
+    "spaces": 4
+  },
+  "lineLength": 120,
+  "multiElementCollectionTrailingCommas": true,
+  "spacesAroundRangeFormationOperators": true
+}
--- a/Package.swift
+++ b/Package.swift
@@ -13,24 +13,42 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.6")
+        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.6"),
     ],
     targets: [
         .executableTarget(
             name: "TransformersCLI",
             dependencies: [
                 "Models", "Generation", "Tokenizers",
-                .product(name: "ArgumentParser", package: "swift-argument-parser")]),
-        .executableTarget(name: "HubCLI", dependencies: ["Hub", .product(name: "ArgumentParser", package: "swift-argument-parser")]),
+                .product(name: "ArgumentParser", package: "swift-argument-parser"),
+            ]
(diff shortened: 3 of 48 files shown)


.swift-format
```diff
@@ -0,0 +1,15 @@
...
+  },
+  "lineLength": 120,
+  "maximumBlankLines": 1,
+  "respectsExistingLineBreaks": true,
+  "lineBreakBeforeEachArgument": true,
+  "multiElementCollectionTrailingCommas": true,
+  "spacesAroundRangeFormationOperators": true,
+  "rules": {
+    "AlwaysUseLowerCamelCase": false
+  }
```
pcuenca, line 14: Not a fun of imposing too many rules given the experimental nature of this project. Something like [what's used in mlx-swift](https://github.com/ml-explore/mlx-swift/blob/main/.swift-format) is enough in my opinion, and it solves most of the distracting differences I've seen in the project so far.

.pre-commit-config.yaml
pcuenca, line 1: I'd remove this file too to make it clear that git hooks are completely optional. Having a CI action to verify style is enough in my opinion. Contributors that want to auto-apply rules are still free to do so.

.github/workflows/swift-format.yml
pcuenca, line 1: Thanks for coming up with this workflow! However, our CI configuration is a bit special for security reasons, we need to add workflows to [this repo](https://github.com/huggingface/hf-workflows/tree/main/.github/workflows). We can work on a separate PR there!

pcuenca (commented):
Thanks a lot for proposing this @DePasqualeOrg! However, I'd prefer to make style rules super minimal at this stage of the project. Can we work on something simpler, like what mlx-swift does?

.swift-format
DePasqualeOrg: I've changed the rules to mirror those used in mlx-swift. This applies a default a line length of 100, which in my opinion makes the code look less nice, as does having multiple function arguments on one line. Not setting `lineBreakBeforeEachArgument` and `multiElementCollectionTrailingCommas` could result in more distracting diffs.

.pre-commit-config.yaml
DePasqualeOrg: I included it because it's included in mlx-swift and mlx-swift-examples, and pre-commit hooks are not required there either. It just makes it simple to run the formatting with `pre-commit run --all-files` instead of having to remember the swift-format command. It would also unify the workflow for contributors to all these packages, so that we just have to remember one command.

.github/workflows/swift-format.yml
DePasqualeOrg: https://github.com/huggingface/hf-workflows/pull/20

.swift-format
pcuenca: Ok, in that case let's use 120 for the line length, as you initially suggested, and `multiElementCollectionTrailingCommas` if you feel strongly about it. But I'd prefer to skip the `lineBreakBeforeEachArgument` for now, I find multi-lines more distracting than helpful when you are browsing the code.

.pre-commit-config.yaml
pcuenca: All right, let's keep it then, I see it being useful for documentation and consistency.

DePasqualeOrg:
Done, thank you!

DePasqualeOrg:
Closed in favor of https://github.com/huggingface/swift-transformers/pull/154

PR #151, opened 2025-01-01
Enable tool use

DePasqualeOrg:
Once https://github.com/johnmai-dev/Jinja/pull/8 gets merged, this will enable tool use.

https://github.com/huggingface/swift-transformers/pull/158 needs to be merged before this PR.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,8 +12,8 @@ let package = Package(
         .executable(name: "hub-cli", targets: ["HubCLI"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/johnmai-dev/Jinja", from: "1.1.0")
+        .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMinor(from: "1.4.0")),
+        .package(url: "https://github.com/johnmai-dev/Jinja", .upToNextMinor(from: "1.1.0"))
     ],
     targets: [
         .executableTarget(
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -399,6 +399,17 @@ public class PreTrainedTokenizer: Tokenizer {
         addGenerationPrompt: Bool = false,
         truncation: Bool = false,
         maxLength: Int? = nil,
+        tools: [ToolSpec]? = nil
+    ) throws -> [Int] {
+        try applyChatTemplate(messages: messages, chatTemplate: chatTemplate, addGenerationPrompt: addGenerationPrompt, truncation: truncation, maxLength: maxLength, tools: tools, additionalContext: nil)
+    }
+
+    public func applyChatTemplate(
+        messages: [Message],
+        chatTemplate: ChatTemplateArgument? = nil,
+        addGenerationPrompt: Bool = false,
+        truncation: Bool = false,
+        maxLength: Int? = nil,
         /// A list of tools (callable functions) that will be accessible to the model. If the template does not
         /// support function calling, this argument will have no effect. Each tool should be passed as a JSON Schema,
         /// giving the name, description and argument types for the tool. See the
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -9,6 +9,9 @@ import Hub
 import Foundation
 import Jinja
 
+public typealias Message = [String: Any]
+public typealias ToolSpec = [String: Any]
+
 enum TokenizerError: Error {
     case missingConfig
     case missingTokenizerClassInConfig
@@ -142,22 +145,26 @@ public protocol Tokenizer {
     var unknownTokenId: Int? { get }
 
     /// The appropriate chat template is selected from the tokenizer config
-    func applyChatTemplate(messages: [[String: String]]) throws -> [Int]
+    func applyChatTemplate(messages: [Message]) throws -> [Int]
+
+    /// The appropriate chat template is selected from the tokenizer config
+    func applyChatTemplate(messages: [Message], tools: [ToolSpec]) throws -> [Int]
 
     /// The chat template is provided as a string literal or specified by name
-    func applyChatTemplate(messages: [[String: String]], chatTemplate: ChatTemplateArgument) throws -> [Int]
+    func applyChatTemplate(messages: [Message], chatTemplate: ChatTemplateArgument) throws -> [Int]
 
     /// The chat template is provided as a string literal
-    func applyChatTemplate(messages: [[String: String]], chatTemplate: String) throws -> [Int]
+    func applyChatTemplate(messages: [Message], chatTemplate: String) throws -> [Int]
 
     func applyChatTemplate(
-        messages: [[String: String]],
+        messages: [Message],
         /// A chat template can optionally be provided or specified by name when several templates are included in the tokenizer config. Normally this is not necessary.
         chatTemplate: ChatTemplateArgument?,
         addGenerationPrompt: Bool,
         truncation: Bool,
         maxLength: Int?,
-        tools: [[String: Any]]?
+        tools: [ToolSpec]?,
+        additionalContext: [String: Any]?
     ) throws -> [Int]
 }
 
@@ -359,20 +366,35 @@ public class PreTrainedTokenizer: Tokenizer {
         model.convertIdToToken(id)
     }
 
-    public func applyChatTemplate(messages: [[String: String]]) throws -> [Int] {
+    public func applyChatTemplate(messages: [Message]) throws -> [Int] {
         try applyChatTemplate(messages: messages, addGenerationPrompt: true)
     }
 
-    public func applyChatTemplate(messages: [[String: String]], chatTemplate: ChatTemplateArgument) throws -> [Int] {
+    public func applyChatTemplate(messages: [Message], tools: [ToolSpec]) throws -> [Int] {
+        try applyChatTemplate(messages: messages, addGenerationPrompt: true, tools: tools)
+    }
+
+    public func applyChatTemplate(messages: [Message], tools: [ToolSpec], additionalContext: [String: Any]) throws
+        -> [Int]
+    {
+        try applyChatTemplate(
+            messages: messages,
+            addGenerationPrompt: true,
+            tools: tools,
+            additionalContext: additionalContext
+        )
+    }
+
+    public func applyChatTemplate(messages: [Message], chatTemplate: ChatTemplateArgument) throws -> [Int] {
         try applyChatTemplate(messages: messages, chatTemplate: chatTemplate, addGenerationPrompt: true)
     }
 
-    public func applyChatTemplate(messages: [[String: String]], chatTemplate: String) throws -> [Int] {
+    public func applyChatTemplate(messages: [Message], chatTemplate: String) throws -> [Int] {
         try applyChatTemplate(messages: messages, chatTemplate: .literal(chatTemplate), addGenerationPrompt: true)
     }
 
     public func applyChatTemplate(
-        messages: [[String: String]],
+        messages: [Message],
         chatTemplate: ChatTemplateArgument? = nil,
         addGenerationPrompt: Bool = false,
         truncation: Bool = false,
@@ -382,8 +404,8 @@ public class PreTrainedTokenizer: Tokenizer {
         /// giving the name, description and argument types for the tool. See the
         /// [chat templating guide](https://huggingface.co/docs/transformers/main/en/chat_templating#automated-function-conversion-for-tool-use)
         /// for more information.
-        /// Note: tool calling is not supported yet, it will be available in a future update.
-        tools: [[String: Any]]? = nil
+        tools: [ToolSpec]? = nil,
+        additionalContext: [String: Any]? = nil
     ) throws -> [Int] {
         var selectedChatTemplate: String?
         if let chatTemplate, case .literal(let template) = chatTemplate {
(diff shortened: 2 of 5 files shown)


pcuenca:
Happy to review this one when it's ready and #150 has been merged.

Package.swift
```diff
@@ -13,24 +13,38 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.6")
+        //        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.6"),
```
johnmai-dev, line 16: This URL needs to be adjusted, change maiqingqiang to johnmai-dev.

https://github.com/johnmai-dev/Jinja

@DePasqualeOrg

DePasqualeOrg:
@pcuenca, this is ready to go as soon as we can merge the update to Swift Jinja as well as the formatting in this repo. I'd love to move forward with this so that we can showcase the new capabilities in mlx-swift-examples.

DePasqualeOrg:
@pcuenca, this is now using the latest release of Swift Jinja, so it's ready to review.

pcuenca:
Can we please remove the formatting changes from this PR 🙏? I'm still going through all the settings in swift-format and will provide feedback on #158.

DePasqualeOrg:
I think that could be difficult, and it might be easier for me to simply re-format the first commit with whatever settings you decide to go with. Or you can merge your own formatting PR and I'll rebase it on that.

I ran formatting first on this PR because formatting differences often cause merge conflicts, and my Xcode uses two spaces for indentation. It's much easier to simply run the formatting before committing, and it's difficult to contribute to a repo that doesn't have consistent formatting.

DePasqualeOrg:
@pcuenca, can we please move forward with this? I have contributed a significant amount of free labor so that people can use the capabilities of the latest models in Swift. There's a lot of interest right now in using DeepSeek R1 in particular. The solution here is simple and doesn't require much effort from you.

DePasqualeOrg:
Here is my proposed path forward, which will make things very easy for everyone:

This repo needs some kind of automated formatting, which other contributors have been asking for since last year. Automated formatting is a part of nearly every major repository that people collaborate on, including nearly all of Hugging Face's public repos, because people have different IDE settings that result in different indentation and other differences.

Automated formatting allows people to keep their preferred IDE settings and run one command to eliminate merge conflicts and unnecessary diffs. It's also very useful when pasting in code from LLMs, which often use different formatting. Running automated formatting allows you to see a clean diff.

As for the specific rules, I will leave that entirely up to @pcuenca's judgement. Even an extremely minimal config with only rules for indentation and line length would already solve the problem.

You can merge your preferred formatting with swift-format (see [my PR](https://github.com/huggingface/swift-transformers/pull/158) for an example), and then I will format, squash, and rebase, and you will see only the changes relevant to this PR in one single commit. That's the beauty of formatting.

pcuenca:
Sure, we can do that, but:

- Swift formatting, in my opinion, has less priority than fixing existing tokenizers or supporting the Deepseek tokenizer. These tasks unblock MLX and fix other downstream use cases. Hence #153, #157, #159, #160, #163 were all handled with urgency. #155 is also super important for the long-term utility of the library, and I'd like to review it as soon as possible.
- Tools support is also more important than agreeing on a syntax. I do appreciate your contribution a lot, as well as all your previous efforts!
- I dislike some of the defaults applied by swift-format and spent some time trying to find a reasonable ground, but couldn't finish yet. I thought that it would only apply the rules that appear in the format file, but it applies several rules by default. I don't take source code changes lightly as they make history more difficult to browse.
- I can't see why cherry-picking the relevant commits from this PR (is it just 24187a222c4d374c51824eb6f502fe2c5ce97e19?) into a separate branch would be complicated. Please, forgive me if I'm wrong, but that would have allowed us to review and test that feature while we work (more slowly) in the swift-format changes.

As I tried to explain offline, swift-format has not been a huge priority so far, but I'll happily adopt it seeing you are contributing regularly to this project and other related ones. However, I don't think it'll be beneficial for the project to hastily adopt the set of syntax changes that were included in the PR.

So my current thinking is to work on the following priorities this week:
- Merge #163. Tests pass and it solves some potential silent issues with the Deepseek tokenizers, so I'm just waiting in case someone else suggests edge cases to try.
- Review, test and merge #155.
- Find reasonable rules for swift-format and finally merge #158.
- Review and merge this PR, as it depends on the previous item.

DePasqualeOrg:
Thanks, I was able to cherry pick the key commit and resolve some conflicts. I didn't think of that possibility before, but I guess that's the beauty of git. Your plan sounds good to me. You can review this PR whenever you're able to.

It's very difficult to test the tool use templates in Swift because of Swift's non-deterministic key order in dictionaries. We're already testing this in Jinja. I added a disabled test as a starting point in case we want to do this later. That would require importing swift-collections only for the tests and using `OrderedDictionary` instead of native Swift dictionaries. I have already tested tool use in https://github.com/ml-explore/mlx-swift-examples/pull/174, and it works as expected.

Tests/TokenizersTests/ChatTemplateTests.swift
```diff
@@ -80,5 +80,66 @@ class ChatTemplateTests: XCTestCase {
...
+//{"name": <function-name>, "arguments": <args-json-object>}
+//</tool_call><|im_end|>
+//<|im_start|>user
+//What is the weather in Paris today?<|im_end|>
+//<|im_start|>assistant
+//
+//"""
+//
+//        XCTAssertEqual(decoded, target)
+//    }
```
pcuenca, line 144: It's a bit cumbersome, but given that you already have the tools dictionary, we could decode the tools section from the decoded string as a dictionary and compare that way:


```swift
    func testQwen2_5WithTools() async throws {
        let tokenizer = try await AutoTokenizer.from(pretrained: "mlx-community/Qwen2.5-7B-Instruct-4bit")

        let weatherQueryMessages: [[String: String]] = [
            [
                "role": "user",
                "content": "What is the weather in Paris today?",
            ]
        ]

        let getCurrentWeatherToolSpec: [String: Any] = [
            "type": "function",
            "function": [
                "name": "get_current_weather",
                "description": "Get the current weather in a given location",
                "parameters": [
                    "type": "object",
                    "properties": [
                        "location": [
                            "type": "string",
                            "description": "The city and state, e.g. San Francisco, CA"
                        ],
                        "unit": [
                            "type": "string",
                            "enum": ["celsius", "fahrenheit"]
                        ]
                    ],
                    "required": ["location"]
                ]
            ]
        ]

        let encoded = try tokenizer.applyChatTemplate(messages: weatherQueryMessages, tools: [getCurrentWeatherToolSpec])
        let decoded = tokenizer.decode(tokens: encoded)

        func assertDictsAreEqual(_ actual: [String: Any], _ expected: [String: Any]) {
            for (key, value) in actual {
                if let nestedDict = value as? [String: Any], let nestedDict2 = expected[key] as? [String: Any] {
                    assertDictsAreEqual(nestedDict, nestedDict2)
                } else if let arrayValue = value as? [String] {
                    let expectedArrayValue = expected[key] as? [String]
                    XCTAssertNotNil(expectedArrayValue)
                    XCTAssertEqual(Set(arrayValue), Set(expectedArrayValue!))
                } else {
                    XCTAssertEqual(value as? String, expected[key] as? String)
                }
            }
        }

        if let startRange = decoded.range(of: "<tools>\n"),
           let endRange = decoded.range(of: "\n</tools>", range: startRange.upperBound..<decoded.endIndex) {
            let toolsSection = String(decoded[startRange.upperBound..<endRange.lowerBound])
            if let toolsDict = try? JSONSerialization.jsonObject(with: toolsSection.data(using: .utf8)!) as? [String : Any] {
                assertDictsAreEqual(toolsDict, getCurrentWeatherToolSpec)
            } else {
                XCTFail("Failed to decode tools section")
            }
        } else {
            XCTFail("Failed to find tools section")
        }
    }
```

I'd also suggest to add additional sanity checks that would be easier to implement: the string starts and ends with the known fragments, or something along those lines.

pcuenca (commented):
Hi @DePasqualeOrg, thanks for the contribution and sorry for the delay in getting to it!

I think the changes are fine, but I think it'd still be useful to have a test. I propose to extract the tools section from the rendered string, as shown in a comment.

Tests/TokenizersTests/ChatTemplateTests.swift
DePasqualeOrg: Thank you! I've added the test as you suggested.

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -142,22 +145,26 @@ public protocol Tokenizer {
...
-        messages: [[String: String]],
+        messages: [Message],
         /// A chat template can optionally be provided or specified by name when several templates are included in the tokenizer config. Normally this is not necessary.
         chatTemplate: ChatTemplateArgument?,
         addGenerationPrompt: Bool,
         truncation: Bool,
         maxLength: Int?,
-        tools: [[String: Any]]?
+        tools: [ToolSpec]?,
+        additionalContext: [String: Any]?
```
pcuenca, line 167: This argument actually introduces a breaking API change for users that provide their own `Tokenizer` implementations. Do you think we could introduce a default implementation somehow? For example, we could keep the previous declaration, and create a default implementation of this method that calls the old one when `additionalContext` is `nil`.

Otherwise we can always introduce this as part of `0.2.0`, but I'd rather add it as an incremental update so most people can use it.

pcuenca (approved):
Looks great, I just realized about a small API breaking change. Would be interested to know. your thoughts.

Sources/Tokenizers/Tokenizer.swift
DePasqualeOrg: Sure, I added an overload for the previous function signature.

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -142,22 +145,26 @@ public protocol Tokenizer {
...
+    /// The appropriate chat template is selected from the tokenizer config
+    func applyChatTemplate(messages: [Message], tools: [ToolSpec]) throws -> [Int]
 
     /// The chat template is provided as a string literal or specified by name
-    func applyChatTemplate(messages: [[String: String]], chatTemplate: ChatTemplateArgument) throws -> [Int]
+    func applyChatTemplate(messages: [Message], chatTemplate: ChatTemplateArgument) throws -> [Int]
 
     /// The chat template is provided as a string literal
-    func applyChatTemplate(messages: [[String: String]], chatTemplate: String) throws -> [Int]
+    func applyChatTemplate(messages: [Message], chatTemplate: String) throws -> [Int]
```
pcuenca, line 157:
```suggestion
    func applyChatTemplate(messages: [Message], chatTemplate: String) throws -> [Int]
    
    func applyChatTemplate(
        messages: [Message],
        /// A chat template can optionally be provided or specified by name when several templates are included in the tokenizer config. Normally this is not necessary.
        chatTemplate: ChatTemplateArgument?,
        addGenerationPrompt: Bool,
        truncation: Bool,
        maxLength: Int?,
        tools: [ToolSpec]?
    ) throws -> [Int]
```

We keep the previous declaration in the protocol.

pcuenca: But we also need something like this:

```
extension Tokenizer {
    /// Call previous signature for backwards compatibility
    func applyChatTemplate(
        messages: [Message],
        /// A chat template can optionally be provided or specified by name when several templates are included in the tokenizer config. Normally this is not necessary.
        chatTemplate: ChatTemplateArgument?,
        addGenerationPrompt: Bool,
        truncation: Bool,
        maxLength: Int?,
        tools: [ToolSpec]?,
        additionalContext: [String: Any]?
    ) throws -> [Int] {
        if additionalContext == nil {
            try applyChatTemplate(messages: messages, chatTemplate: chatTemplate, addGenerationPrompt: addGenerationPrompt, truncation: truncation, maxLength: maxLength, tools: tools)
        } else {
            throw TokenizerError.chatTemplate("Not implemented")
        }
    }
}
```

It's a bit confusing, but the idea is: if you have written your own `Tokenizer` implementation in your project, you will have implemented your own version of `applyChatTemplate` with the previous signature. But now there's a new protocol requirement that takes a new argument (the `additionalContext`) that does not exist in your code, because you implemented the old version. So your code will not compile. With the extension above, we call the "old" implementation, so the code will still compile and work.

I don't think this affects our implementation in `PreTrainedTokenizer` below, everything should just keep working.

What do you think?

DePasqualeOrg: I'm sorry, I don't really understand what the problem is or what the solution should be. Would you like to make the change yourself?

DePasqualeOrg: I think it would be easier if you make the change yourself. I'm not sure if I fully understand the problem.

pcuenca:
> it would be easier if you make the change yourself

I can't. Can you check if you [allowed "edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)?

pcuenca:
Otherwise we can merge and I'll open a followup pr.

DePasqualeOrg:
"Allow edits and access to secrets by maintainers" is checked.

pcuenca:
Ah my bad, it worked now. Let's wait for the CI to clear 🤞

pcuenca:
Merging, I hope I didn't break anything! Thanks @DePasqualeOrg!

DePasqualeOrg:
Thank you! Whenever you're ready, could you create a new release tag so that this can get picked up in my PRs for tool use and vision chat templates in mlx-swift-examples?

PR #152, opened 2025-01-08
Fix tokenizer for Phi 4

DePasqualeOrg:
Phi 4 is not behaving correctly, as you can see from this debug output:

```
::: Prompt rendered with template:
<|im_start|>user<|im_sep|>What is the difference between a fruit and a vegetable?<|im_end|><|im_start|>assistant<|im_sep|>

::: Tokenized text: ["<|im_start|>", "<|im_sep|>", "<|im_end|>", "<|im_start|>", "<|im_sep|>"]

::: Prompt tokens decoded:
<|im_start|><|im_sep|><|im_end|><|im_start|><|im_sep|>
```

Demo with LLMEval here: https://github.com/DePasqualeOrg/mlx-swift-examples/tree/test-phi-4

@pcuenca, it looks like the tokenizer is only processing the special tokens and ignoring the content between them. Do you have any idea why this might be happening?

### DIFF
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -313,7 +313,9 @@ public class PreTrainedTokenizer: Tokenizer {
 
     /// Main entry point
     public func encode(text: String, addSpecialTokens: Bool = true) -> [Int] {
-        return postProcess(tokenize(text: text), addSpecialTokens: addSpecialTokens).map { model.convertTokenToId($0)! }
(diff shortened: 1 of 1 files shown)


pcuenca:
Taking a look

pcuenca:
I think the `SplitTokenizer` does not yet support the `Removed` behaviour, which is used by this tokenizer.

pcuenca:
Working on #153

pcuenca:
Superseded by #153 (I attributed you as a co-author :))

PR #153, opened 2025-01-09
Fix Phi 4 tokenization

pcuenca:


Tests/TokenizersTests/TokenizerTests.swift
```diff
@@ -95,6 +95,17 @@ class GemmaUnicodeTests: XCTestCase {
     }
 }
 
+class PhiSimpleTests: XCTestCase {
```
pcuenca, line 98: This test passes but [these ones](https://github.com/huggingface/swift-transformers/blob/43ae46d22a8870d8034129972d4f19dcdbf5ead9/Tests/PreTokenizerTests/PreTokenizerTests.swift#L139) do not. I don't know if these tests were wrong all along or there's something else going on.

pcuenca:
Tests pass, merging. Thanks a lot @DePasqualeOrg!

DePasqualeOrg:
Thank you @pcuenca!

PR #154, opened 2025-01-10
Format with swift-format

DePasqualeOrg:
I've removed the GitHub action while we wait for https://github.com/huggingface/hf-workflows/pull/20 to be approved. It would be helpful to format the code now to reduce friction with https://github.com/huggingface/swift-transformers/pull/151.


DePasqualeOrg:
Something has gotten mixed up with the latest changes after I opened this, and I think I now have to open a new PR.

PR #155, opened 2025-01-15
Add metadata support with tests

ardaatahan:
This PR aims to bring over some of huggingface_hub Python library's capabilities to swift-transformers:

- Adds file metadata support to prevent unnecessary redownloads, ensure file integrity, and update to the latest file versions when needed.
- Disables automatic redirection during head requests to fix getFileMetadata, which currently doesn't return expected commit hash and etag for lfs files.
- Adds additional tests to verify metadata handling and download logic.

### DIFF
--- a/Sources/Hub/Downloader.swift
+++ b/Sources/Hub/Downloader.swift
@@ -37,12 +37,7 @@ class Downloader: NSObject, ObservableObject {
         to destination: URL,
         metadataDirURL: URL,
         using authToken: String? = nil,
-        inBackground: Bool = false,
-        resumeSize: Int = 0,
-        headers: [String: String]? = nil,
-        expectedSize: Int? = nil,
-        timeout: TimeInterval = 10,
-        numRetries: Int = 5
+        inBackground: Bool = false
     ) {
         self.destination = destination
         let filename = (destination.lastPathComponent as NSString).deletingPathExtension
@@ -59,18 +54,10 @@ class Downloader: NSObject, ObservableObject {
 
         self.urlSession = URLSession(configuration: config, delegate: self, delegateQueue: nil)
 
-        setupDownload(from: url, with: authToken, resumeSize: resumeSize, headers: headers, expectedSize: expectedSize, timeout: timeout, numRetries: numRetries)
+        setupDownload(from: url, with: authToken)
     }
 
-    private func setupDownload(
-        from url: URL,
-        with authToken: String?,
-        resumeSize: Int,
-        headers: [String: String]?,
-        expectedSize: Int?,
-        timeout: TimeInterval,
-        numRetries: Int
-    ) {
+    private func setupDownload(from url: URL, with authToken: String?) {
         downloadState.value = .downloading(0)
         urlSession?.getAllTasks { tasks in
             // If there's an existing pending background task with the same URL, let it proceed.
@@ -95,99 +82,14 @@ class Downloader: NSObject, ObservableObject {
                 }
             }
             var request = URLRequest(url: url)
-            var requestHeaders = headers ?? [:]
             if let authToken = authToken {
-                requestHeaders["Authorization"] = "Bearer \(authToken)"
+                request.setValue("Bearer \(authToken)", forHTTPHeaderField: "Authorization")
             }
-            if resumeSize > 0 {
-                requestHeaders["Range"] = "bytes=\(resumeSize)-"
-            }
-            request.timeoutInterval = timeout
-            request.allHTTPHeaderFields = requestHeaders
 
-            Task {
-                do {
-                    try await self.downloadWithStreaming(request: request, resumeSize: resumeSize, numRetries: numRetries, expectedSize: expectedSize)
-                } catch {
-                    self.downloadState.value = .failed(error)
-                }
-            }
+            self.urlSession?.downloadTask(with: request).resume()
         }
     }
 
-    private func downloadWithStreaming(
-        request: URLRequest,
-        resumeSize: Int,
-        numRetries: Int,
-        expectedSize: Int?
-    ) async throws {
-        guard let session = self.urlSession else {
-            throw DownloadError.unexpectedError
-        }
-        let tempURL = FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString)
-        FileManager.default.createFile(atPath: tempURL.path, contents: nil)
-        let tempFile = try FileHandle(forWritingTo: tempURL)
-        
-        defer { tempFile.closeFile() }
-        
-        let (asyncBytes, response) = try await session.bytes(for: request)
-        guard let response = response as? HTTPURLResponse else {
-            throw DownloadError.unexpectedError
-        }
-        
-        guard (200..<300).contains(response.statusCode) else {
-            throw DownloadError.unexpectedError
-        }
-        
-        var downloadedSize = resumeSize
-        
-        var buffer = Data(capacity: chunkSize)
-        var newNumRetries = numRetries
-        
-        do {
-            for try await byte in asyncBytes {
-                buffer.append(byte)
-                if buffer.count == chunkSize {
-                    if !buffer.isEmpty { // Filter out keep-alive chunks
-                        try tempFile.write(contentsOf: buffer)
-                        buffer.removeAll(keepingCapacity: true)
-                        downloadedSize += chunkSize
-                        newNumRetries = 5
-                    }
-                }
-            }
-            
-            if !buffer.isEmpty {
-                try tempFile.write(contentsOf: buffer)
-                downloadedSize += buffer.count
-                buffer.removeAll(keepingCapacity: true)
-                newNumRetries = 5
-            }
-        } catch let error as URLError {
-            if newNumRetries <= 0 {
-                throw error
-            }
-            try await Task.sleep(nanoseconds: 1_000_000_000)
-            
-            try await downloadWithStreaming(
-                request: request,
-                resumeSize: downloadedSize,
-                numRetries: newNumRetries - 1,
-                expectedSize: expectedSize
-            )
-        }
-        
-        let actualSize = try tempFile.seekToEnd()
-        if let expectedSize = expectedSize, expectedSize != actualSize {
-            throw DownloadError.unexpectedError
-        }
-        
-        tempFile.closeFile()
-        try FileManager.default.moveDownloadedFile(from: tempURL, to: destination)
-        
-        downloadState.value = .completed(destination)
-    }
-    
     @discardableResult
     func waitUntilDone() throws -> URL {
         // It's either this, or stream the bytes ourselves (add to a buffer, save to disk, etc; boring and finicky)
--- a/Sources/Hub/Downloader.swift
+++ b/Sources/Hub/Downloader.swift
@@ -11,6 +11,9 @@ import Combine
 
 class Downloader: NSObject, ObservableObject {
     private(set) var destination: URL
+    private(set) var metadataDestination: URL
+
+    private let chunkSize = 10 * 1024 * 1024  // 10MB
 
     enum DownloadState {
         case notStarted
@@ -29,8 +32,21 @@ class Downloader: NSObject, ObservableObject {
 
     private var urlSession: URLSession? = nil
 
-    init(from url: URL, to destination: URL, using authToken: String? = nil, inBackground: Bool = false) {
+    init(
+        from url: URL,
+        to destination: URL,
+        metadataDirURL: URL,
+        using authToken: String? = nil,
+        inBackground: Bool = false,
+        resumeSize: Int = 0,
+        headers: [String: String]? = nil,
+        expectedSize: Int? = nil,
+        timeout: TimeInterval = 10,
+        numRetries: Int = 5
+    ) {
         self.destination = destination
+        let filename = (destination.lastPathComponent as NSString).deletingPathExtension
+        self.metadataDestination = metadataDirURL.appending(component: "\(filename).metadata")
         super.init()
         let sessionIdentifier = "swift-transformers.hub.downloader"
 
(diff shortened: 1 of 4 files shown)


Sources/Hub/Downloader.swift
```diff
@@ -11,6 +11,9 @@ import Combine
 
 class Downloader: NSObject, ObservableObject {
     private(set) var destination: URL
+    private(set) var metadataDestination: URL
```
pcuenca, line 14: Is this used by the Downloader at all, or is it a remnant of deleted commits?

Sources/Hub/HubApi.swift
```diff
@@ -138,6 +148,26 @@ public extension HubApi {
     }
 }
 
+/// Additional Errors
+public extension HubApi {
+    enum EnvironmentError: LocalizedError {
+        case consistencyError(String)
+        case diskSpaceError(String)
+        case permissionError(String)
```
pcuenca, line 156: I think these are unused

Sources/Hub/HubApi.swift
```diff
@@ -201,6 +231,12 @@ public extension HubApi {
             repoDestination.appending(path: relativeFilename)
         }
         
+        var metadataDestination: URL {
+            repoDestination
+                .appendingPathComponent(".cache")
+                .appendingPathComponent("huggingface")
```
pcuenca, line 237:
```suggestion
                .appendingPathComponent("huggingface")
                .appendingPathComponent("download")
```

Maybe we should consider adding this, for parity with the hub cli

Sources/Hub/HubApi.swift
```diff
@@ -209,16 +245,160 @@ public extension HubApi {
...
+                    }
+                    
+                    let commitHash = lines[0].trimmingCharacters(in: .whitespacesAndNewlines)
+                    let etag = lines[1].trimmingCharacters(in: .whitespacesAndNewlines)
+                    guard let timestamp = Double(lines[2].trimmingCharacters(in: .whitespacesAndNewlines)) else {
+                        throw EnvironmentError.invalidMetadataError("Missing or invalid timestamp.")
+                    }
+                    let timestampDate = Date(timeIntervalSince1970: timestamp)
+                            
+                    // TODO: check if file hasn't been modified since the metadata was saved
```
pcuenca, line 280: is this still relevant?

Sources/Hub/HubApi.swift
```diff
@@ -209,16 +245,160 @@ public extension HubApi {
...
+                    return nil
+                }
+            }
+                
+            // metadata file does not exist
+            return nil
+        }
+        
+        func isValidSHA256(_ hash: String) -> Bool {
+            let sha256Pattern = "^[0-9a-f]{64}$"
```
pcuenca, line 299:
```suggestion
            let sha256Pattern = "^[0-9a-f]{40}$"
```

I think?

Sources/Hub/HubApi.swift
```diff
@@ -209,16 +245,160 @@ public extension HubApi {
...
+            return regex?.firstMatch(in: hash, options: [], range: range) != nil
+        }
+        
+        /// Reference: https://github.com/huggingface/huggingface_hub/blob/b2c9a148d465b43ab90fab6e4ebcbbf5a9df27d4/src/huggingface_hub/_local_folder.py#L391
+        func writeDownloadMetadata(commitHash: String, etag: String, metadataRelativePath: String) throws {
+            let metadataContent = "\(commitHash)\n\(etag)\n\(Date().timeIntervalSince1970)\n"
+            let metadataPath = metadataDestination.appending(component: metadataRelativePath)
+            
+            do {
+                try FileManager.default.createDirectory(at: metadataPath.deletingLastPathComponent(), withIntermediateDirectories: true)
```
pcuenca, line 311: Is this the same path that was created in `prepareMetadataDestination`? Or potentially a subdirectory?

Sources/Hub/HubApi.swift
```diff
@@ -209,16 +245,160 @@ public extension HubApi {
...
+        
+        
         // Note we go from Combine in Downloader to callback-based progress reporting
         // We'll probably need to support Combine as well to play well with Swift UI
         // (See for example PipelineLoader in swift-coreml-diffusers)
         @discardableResult
         func download(progressHandler: @escaping (Double) -> Void) async throws -> URL {
-            guard !downloaded else { return destination }
-
+            var metadataRelativePath = (relativeFilename as NSString).deletingPathExtension
```
pcuenca, line 355: I think we should keep the extension here to:
- Avoid silent overrides
- Keep parity with the Hub CLI

Tests/HubTests/HubApiTests.swift
```diff
@@ -252,4 +272,239 @@ class SnapshotDownloadTests: XCTestCase {
...
+            print("Total Progress: \(progress.fractionCompleted)")
+            print("Files Completed: \(progress.completedUnitCount) of \(progress.totalUnitCount)")
+            lastProgress = progress
+        }
+        
+        attributes = try FileManager.default.attributesOfItem(atPath: configPath.path)
+        let secondDownloadTimestamp = attributes[.modificationDate] as! Date
+
+        // File will not be downloaded again thus last modified date will remain unchanged
+        XCTAssertTrue(originalTimestamp == secondDownloadTimestamp)
```
pcuenca, line 368: Nice test! I'd add a check (or an additional test) about the metadata files being identical the second time the download is performed. This would ensure the short path in https://github.com/huggingface/swift-transformers/pull/155/files#diff-d008547ca96c12875517345e1c3ddd2cdae19582c8f414b275e8b29f10bf801fR364 was correctly followed.

pcuenca (commented):
Looks great, and sorry for the delay in getting to this PR 🙏 

I think there are a couple of inconsistencies that should be straightforward to fix. I also suggested an additional test to verify metadata reuse.

More generally, I think we should exactly mimic the cache structure of the Hub CLI because we're almost there anyway, and it could potentially be useful down the line.

Great work, thanks a lot! 🙌

Sources/Hub/HubApi.swift
```diff
@@ -209,16 +245,160 @@ public extension HubApi {
...
+        func prepareMetadataDestination() throws {
+            try FileManager.default.createDirectory(at: metadataDestination, withIntermediateDirectories: true, attributes: nil)
+        }
+        
+        /// Reads metadata about a file in the local directory related to a download process.
+        ///
+        /// Reference: https://github.com/huggingface/huggingface_hub/blob/b2c9a148d465b43ab90fab6e4ebcbbf5a9df27d4/src/huggingface_hub/_local_folder.py#L263
+        ///
+        /// - Parameters:
+        ///   - localDir: The local directory where files are downloaded.
```
pcuenca, line 258:
```suggestion
        ///   - localDir: The local directory where metadata files are downloaded.
```

Tests/HubTests/HubApiTests.swift
```diff
@@ -144,6 +144,26 @@ class HubApiTests: XCTestCase {
             XCTAssertGreaterThan(metadata.size!, 0)
         }
     }
+    
+    /// Verify with `curl -I https://huggingface.co/coreml-projects/Llama-2-7b-chat-coreml/resolve/main/llama-2-7b-chat.mlpackage/Data/com.apple.CoreML/model.mlmodel`
+    func testGetLargeFileMetadata() async throws {
+        do {
+            let revision = "eaf97358a37d03fd48e5a87d15aff2e8423c1afb"
+            let etag = "fc329090bfbb2570382c9af997cffd5f4b78b39b8aeca62076db69534e020107"
+            let location = "https://cdn-lfs.hf.co/repos/4a/4e/4a4e587f66a2979dcd75e1d7324df8ee9ef74be3582a05bea31c2c26d0d467d0/fc329090bfbb2570382c9af997cffd5f4b78b39b8aeca62076db69534e020107?response-content-disposition=inline%3B+filename*%3DUTF-8%27%27model.mlmodel%3B+filename%3D%22model.mlmodel"
```
pcuenca, line 153: I'd also add tests that exercise the full download cycle of LFS objects. For example, we can use [this small repo](https://huggingface.co/pcuenq/smol-lfs/tree/main).

Sources/Hub/Downloader.swift
ardaatahan: remnant of deleted commits, just reverted this to the original

Sources/Hub/HubApi.swift
ardaatahan: yeah this is implemented in the huggingface_hub function: https://github.com/huggingface/huggingface_hub/blob/2fdc6f48ef5e6b22ee9bcdc1945948ac070da675/src/huggingface_hub/_local_folder.py#L265

Sources/Hub/HubApi.swift
ardaatahan: 64 should be correct. Reference: https://github.com/huggingface/huggingface_hub/blob/2fdc6f48ef5e6b22ee9bcdc1945948ac070da675/src/huggingface_hub/file_download.py#L81

Sources/Hub/HubApi.swift
ardaatahan: Potentially a subdirectory.

It's basically following the directory structure of the downloaded files here as well to make sure there's no silent overrides in case there're several files with the same name under different directories.

Tests/HubTests/HubApiTests.swift
ardaatahan: Added two additional test cases that download lfs files.

ardaatahan:
Thanks for the review!

I'm considering adding offline mode support under HubApi, controlled by a boolean flag in the constructor. If the flag isn’t provided, it could default to using something like isExpensive as a fallback.

This would allow us to return already downloaded files after verification without requiring a new download when on cellular data.

Curious to hear your thoughts!

Sources/Hub/HubApi.swift
pcuenca: Let's maybe link to the `stat` line if you want.

Tests/HubTests/HubApiTests.swift
```diff
@@ -167,7 +167,7 @@ class HubApiTests: XCTestCase {
 }
 
 class SnapshotDownloadTests: XCTestCase {
-    let repo = "coreml-projects/Llama-2-7b-chat-coreml"
+    var repo = "coreml-projects/Llama-2-7b-chat-coreml"
```
pcuenca, line 170: Let's move this back to `let` and update `testDownloadSmolLargeFile` to define a new ivar.

Sources/Hub/HubApi.swift
pcuenca: Ah, you are right, but that's for SHA256 etags that represent LFS files. Commit hashes will never match this, we have to use [this pattern](https://github.com/huggingface/huggingface_hub/blob/07e1adb1b1be2eb8cc27520dabf619444af64efb/src/huggingface_hub/file_download.py#L78) instead.

Sources/Hub/HubApi.swift
```diff
@@ -209,15 +240,158 @@ public extension HubApi {
...
+            let metadataRelativePath = "\(relativeFilename).metadata"
+                        
+            let localMetadata = try readDownloadMetadata(localDir: metadataDestination, filePath: metadataRelativePath)
+            let remoteMetadata = try await HubApi.shared.getFileMetadata(url: source)
+            
+            let localCommitHash = localMetadata?.commitHash ?? ""
+            let remoteCommitHash = remoteMetadata.commitHash ?? ""
+            
+            // Local file exists + metadata exists + commit_hash matches => return file
+            if isValidSHA256(remoteCommitHash) && downloaded && localMetadata != nil && localCommitHash == remoteCommitHash {
```
pcuenca, line 359: The `isValidSHA256` check will always fail here on the `remoteCommitHash`.

Sources/Hub/HubApi.swift
```diff
@@ -209,15 +240,158 @@ public extension HubApi {
...
+                if localMetadata?.etag == remoteEtag {
+                    try writeDownloadMetadata(commitHash: remoteCommitHash, etag: remoteEtag, metadataRelativePath: metadataRelativePath)
+                    return destination
+                }
+                
+                // metadata is outdated + etag is a sha256
+                // => means it's an LFS file (large)
+                // => let's compute local hash and compare
+                // => if match, update metadata and return file
+                if localMetadata != nil && isValidSHA256(remoteEtag) {
```
pcuenca, line 382: This `isValidSHA256` will succeed when it's an LFS file.

pcuenca (commented):
I think there's a confusion with the hash validation of the commit hashes versus the etags of LFS files.

I would suggest to:
- Add another validation function for commit hashes (40 chars)
- Write a test (if possible, it may be convoluted) that retrieves metadata from an LFS file and verifies that:
  * The commit hash matches the commit validation function
  * The etag hash matches the isValidSHA256 function

pcuenca:
> I'm considering adding offline mode support under HubApi, controlled by a boolean flag in the constructor. If the flag isn’t provided, it could default to using something like isExpensive as a fallback.
> 
> This would allow us to return already downloaded files after verification without requiring a new download when on cellular data.

Sure, sounds great! Maybe get this merged first and then open a new PR, to make things simpler?

ardaatahan:
> > I'm considering adding offline mode support under HubApi, controlled by a boolean flag in the constructor. If the flag isn’t provided, it could default to using something like isExpensive as a fallback.
> > This would allow us to return already downloaded files after verification without requiring a new download when on cellular data.
> 
> Sure, sounds great! Maybe get this merged first and then open a new PR, to make things simpler?

Sounds good!

ardaatahan:
On a different note: after these changes get merged, calling snapshot will trigger a lot of redownloads on mobile devices due to the absence of existing metadata files. This is expected behavior, but please be mindful of the potential impact on bandwidth and storage usage.

pcuenca (approved):

pcuenca:
> calling snapshot will trigger a lot of redownloads

ah, you are right. Do you think we should add the offline mode first then? And/or tag this as a new version `0.2.0`.

ardaatahan:
> ah, you are right. Do you think we should add the offline mode first then? And/or tag this as a new version `0.2.0`.

I'm curious about your opinion on changing the download logic to first download metadata files if they're missing (and redownload non-lfs files), then only redownload lfs files if their checksums don't match the etag or if they're missing. This could safeguard against high bandwidth usage without enabling offline mode, which would completely restrict any new downloads

pcuenca:
That sounds like a great idea!

ZachNagengast:
With these changes, full LFS files will still potentially be downloaded when calling snapshot, but only if they don't match the remote hash, which is good because the local model is stale in that case. I think this is safe to merge before offline mode since current behavior of snapshot is already expected to make a decision on downloading the full file internally. 

Once offline mode is implemented, devs will be able to safely call snapshot everytime with an extra offline mode parameter that simply checks if the local files are valid based on the stored metadata. This is useful for the case that a device is on an expensive or limited cell connection, and devs want to make a decision (or even prompt the user) if it is ok to download in this state, or just continue to use a stale model in offline mode until on a better connection.

pcuenca (approved):
Thanks for the great contribution!

PR #157, opened 2025-01-16
Bert fixes

### LINKED ISSUE #156: BertTokenizer outputs differ between cased and uncased pretrained tokenizers
On my fork, I’ve introduced support for DistilBERT models and the ability to instantiate tokenizers from local files. These modifications are tailored to my specific needs, as I primarily work with the Polish language. I aim to contribute these changes to the mainline library, however I do encounter problems with the foundational codebase. Input from the core team would be essential for proceeding.

## Objectives

The key goal is to ensure consistent tokenizer behaviour between training (Python) and inference (iOS), particularly being able to load tokenizer from the filesystem and handle Polish diacritics like “ą”, “ę”, etc.

## Diagnosis Summary

**Edit**
- `BasicTokenizer` does not act in the same way as in reference implementation.
- `do_lower_case` and `strip_accents` are not taken into account 

### Swift 
```swift
let tokenizer = try await AutoTokenizer.from(pretrained: "google-bert/bert-base-cased") as? PreTrainedTokenizer else {
tokenizer.encode(text: "mąka")
``` 

gives:
```
[101, 12477, 1968, 102]
```

### Python 

```python
tokenizer1 = transformers.AutoTokenizer.from_pretrained("google-bert/bert-base-cased")
print(tokenizer1("mąka"))
```

gives:
```
[101, 182, 28213, 1968, 102]
```

## Proposals

To overcome these limitations, consider at least one of the following:
- ~~Expose `strip_accents` in the public API of Transformers.~~
- ~~Implement a Swift variant of the `save_pretrained` function that handles nil values gracefully.~~
- `BasicTokenizer` become aware of `do_lower_case` config option.

Dropped a massive update. The original post got quickly outdated to the point it made no sense anymore.

pcuenca:
Fixes #156


Sources/Tokenizers/Normalizer.swift
```diff
@@ -146,14 +146,14 @@ class NFKCNormalizer: Normalizer {
...
-    let shouldStripAccents: Bool?
+    let shouldStripAccents: Bool
     let shouldLowercase: Bool
 
     required init(config: Config) {
         self.shouldCleanText = config.cleanText?.boolValue ?? true
         self.shouldHandleChineseChars = config.handleChineseChars?.boolValue ?? true
-        self.shouldStripAccents = config.stripAccents?.boolValue
         self.shouldLowercase = config.lowercase?.boolValue ?? true
+        self.shouldStripAccents = config.stripAccents?.boolValue ?? shouldLowercase
```
pcuenca, line 156: see https://docs.rs/tokenizers/latest/src/tokenizers/normalizers/bert.rs.html#119-137

Sources/Tokenizers/Normalizer.swift
```diff
@@ -219,12 +219,10 @@ class BertNormalizer: Normalizer {
...
-        text.decomposedStringWithCanonicalMapping
-            .filter {
-                $0.unicodeScalars.allSatisfy { scalar in
-                    !(0x0300 <= scalar.value && scalar.value <= 0x036F)
-                }
-            }
+        // This might be the same as `text.folding(options: .diacriticInsensitive, locale: nil)`
+        String(text.decomposedStringWithCanonicalMapping.unicodeScalars.filter { scalar in
+            !(0x0300 <= scalar.value && scalar.value <= 0x036F)
+        })
```
pcuenca, line 225: This was one of the main problems - stripAccents was removing accented chars, but we should only remove the diacritic marks.

pcuenca:
https://github.com/huggingface/swift-transformers/issues/156#issuecomment-2598110390

pcuenca:
Thanks for reporting #156, @piotrkowalczuk! There were several inconsistencies that should be resolved now, and I added much more comprehensive tests. I double-checked the test cases against the Python and Rust implementations, and some of the most confusing ones [against the original codebase](https://github.com/google-research/bert/blob/master/tokenization.py).

Note there's still an inconsistency with respect to the "fast" Rust-based transformers tokenizers, when decoding tokenized data:

Fast tokenizer:
```py
>>> from transformers import AutoTokenizer
>>> t = AutoTokenizer.from_pretrained("google-bert/bert-base-uncased")
>>> t.decode(t.encode("l'eure"), skip_special_tokens=True)
"l ' eure"
```

Slow (Python based) tokenizer:
```py
>>> t = AutoTokenizer.from_pretrained("google-bert/bert-base-uncased", use_fast=False)
>>> t.decode(t.encode("l'eure"), skip_special_tokens=True)
"l'eure"
```

We believe the "slow" interpretation to be correct, and it's the same approach @xenova took for transformers.js. There's no decoding code in the [reference implementation](https://github.com/google-research/bert/blob/master/tokenization.py), but I believe the intention was to strip those characters.

--- 

Could you please give it a final look / test drive before merging @piotrkowalczuk?

cc @Vaibhavs10 because this was fun.

piotrkowalczuk:
LGTM. I was able to integrate your branch into my app. All the regression cases worked fine.  Thanks @pcuenca.

pcuenca:
Cool, merging then!

PR #158, opened 2025-01-19
Format with swift-format

DePasqualeOrg:
@pcuenca, can we please merge this, since every time more commits get added, I have to update it?


DePasqualeOrg:
@pcuenca, if you're worried about the extent of the changes from the formatting, you can remove the last commit and run the formatting yourself. It should only take a minute to take care of this, and the longer it gets left open, the more problems it creates down the line for  https://github.com/huggingface/swift-transformers/pull/151, which is ready to go and only waiting on this.

FL33TW00D:
@DePasqualeOrg Thanks for this, I think this will be superseded by #174

PR #159, opened 2025-01-24
Deepseek R1 tokenization support

pcuenca:
The new test does not pass for some reason. Is there anything you think I'm doing wrong @DePasqualeOrg?

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.6")
+        .package(url: "https://github.com/johnmai-dev/Jinja", from: "1.1.0")
     ],
     targets: [
         .executableTarget(
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/johnmai-dev/Jinja", from: "1.1.0")
+        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.6")
     ],
     targets: [
         .executableTarget(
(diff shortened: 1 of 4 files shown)


DePasqualeOrg:
The encoded tokens from the test look like this:

```
[4913, 70398, 788, 895, 11, 13265, 1313, 788, 330, 19337, 3323, 497, 330, 38460, 788, 830, 11, 330, 75, 13105, 788, 895, 11, 330, 1796, 788, 330, 151646, 497, 330, 15338, 13533, 788, 895, 92, 151644, 74785, 279, 23670, 15473, 4128, 13, 151645]
```

And the test target looks like this:

```
[151646, 151644, 74785, 279, 23670, 15473, 4128, 13, 151645]
```

It's crashing when trying to decode the test target because of the force unwrapping in `decode` in `ByteLevelDecoder`. I'm guessing that the test target is not correct and some tokens don't exist in the vocabulary.

pcuenca:
The test targets were obtained from the Python tokenizer, they correspond to `<｜begin▁of▁sentence｜><｜User｜>Describe the Swift programming language.<｜Assistant｜>`. The problem here is that the `bos_token` is passed in the context as a dictionary, not a String. This means that the result from `applyChatTemplate` won't be correct, as the test shows.

Looking into it.

DePasqualeOrg:
As you can see [here](https://github.com/DePasqualeOrg/mlx-swift-examples/tree/test-deepseek-r1), the prompt from the test is being encoded correctly, and there are no problems interacting with the model, but as soon as you call `decode` on the encoded tokens, it crashes. I think the problem must be somewhere in `swift-transformers`. Perhaps it's a text encoding issue. I noticed that spaces are getting encoded as an unusual character.

@pcuenca, since you're more familiar with the library's internals than me, perhaps you have a better intuition about how to approach the solution. My initial attempts at solutions with Sonnet and the entire library as context were unsuccessful.

Package.swift
```diff
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.4.0"),
-        .package(url: "https://github.com/maiqingqiang/Jinja", from: "1.0.6")
+        .package(url: "https://github.com/johnmai-dev/Jinja", from: "1.1.0")
```
pcuenca, line 16: I'm going to move this and the chat template test to a new PR, since the rest of the fixes here are more general and unrelated to the new jinja engine.

PR #163, opened 2025-01-26
Override Llama tokenizer post-processor if necessary

### LINKED ISSUE #161: Deepseek / Llama tokenization mismatch
While working on the Deepseek tokenization fixes, I found this inconsistency:

Python
```python
from transformers import AutoTokenizer

t = AutoTokenizer.from_pretrained("deepseek-ai/DeepSeek-R1-Distill-Qwen-7B")
t.encode("Who are you?")
# [151646, 15191, 525, 498, 30]
```

The first token, `151646`, is the bos token.

However, in swift-transformers, this fails:
```swift
    func testDeepSeek() async throws {
        guard let tokenizer = try await AutoTokenizer.from(pretrained: "deepseek-ai/DeepSeek-R1-Distill-Qwen-7B") as? PreTrainedTokenizer else {
            XCTFail()
            return
        }
        XCTAssertEqual(tokenizer.encode(text: "Who are you?"), [151646, 15191, 525, 498, 30])
    }
```

The bos token is not prepended to the output.

I think the reason is [this function in transformers](https://github.com/huggingface/transformers/blob/bcb841f0073fcd7a4fb88ea8064313c17dcab04a/src/transformers/models/llama/tokenization_llama_fast.py#L181) that updates the post-processor in the Llama Tokenizers. A template post-processor is added on tokenizer load.

I tested with a Llama tokenizer and it works, because the tokenizer was pushed to the Hub with the template post-processor. However, this is not the case for the Deepseek tokenizers I've seen. We could open PRs to the repos in the Hub, but given all the deepseek activity there could be many derivative repos that reuse the same tokenizer.

Therefore, it might make sense to reproduce the transformers hack in swift-transformers.

pcuenca:
See https://github.com/huggingface/transformers/blob/bcb841f0073fcd7a4fb88ea8064313c17dcab04a/src/transformers/models/llama/tokenization_llama_fast.py#L181 for reference

Fixes #161

### DIFF
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -17,6 +17,7 @@ enum TokenizerError: Error {
     case malformedVocab
     case chatTemplate(String)
     case tooLong(String)
+    case mismatchedConfig(String)
 }
 
 public protocol TokenizingModel {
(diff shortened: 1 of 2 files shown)


pcuenca:
cc @xenova, I'll be merging this in some hours.

PR #168, opened 2025-01-30
Experimental: tokenizers with and without templates

pcuenca:
Pending stuff / edge cases / annoying issues

- [ ] Strict concurrency errors and warnings. Fix and verify they still work for Swift 5.9
- [ ] When compiling the package in Xcode, we get the availability warnings because `Tokenizers` is built without the trait. It works in the CLI with `swift build --traits ChatTemplates`. I don't know if there's a workaround for Xcode.
- [ ] Testing builds `Tokenizers` without the trait, so the chat template tests don't pass. It works in the command-line with `swift test --traits ChatTemplates` (or `swift test --filter TokenizersTests.ChatTemplateTests --traits ChatTemplates` to run just the chat template tests).

----

This builds on top of #166 by @greenrazer.

Two new top-level library products are exposed:

- `Hub`
- `Tokenizers`

(`Transformers` still exists, and comprises everything in the package including tensor ops and Core ML inference).

The `Tokenizers` library is heavy when using chat templates, because it requires a Jinja template engine and swift-collections. Thanks to @mattt, we can easily opt-in to using this feature using package traits, which require Swift 6.1. We attempted another solution that was compatible with previous versions of Swift, but we found it to be too unreliable.

How to use `Tokenizers`.

* On Swift < 6.1, chat templates (and the corresponding dependencies) are always available. `Package@swift-5.9.swift` applies.
* On Swift > 6.1
  - **Without** chat templates, declare the dependency as usual:
    ```swift
    dependencies: [
        .package(url: "https://github.com/huggingface/swift-transformers.git", branch: "hub-tokenizers-templates"),
    ],
    ```
  - **Opt-in to chat templates** using the `ChatTemplates` trait:
    ```swift
    dependencies: [
        .package(
            url: "https://github.com/huggingface/swift-transformers.git",
            branch: "hub-tokenizers-templates",
            traits: ["ChatTemplates"],
        ),
    ],
    ```

----

<details>
 <summary>Previous discussion for reference (no longer applies)</summary>

The goal is to be able to opt-in to the chat template feature, which carries the Jinja dependency, which in turn depends on swift-collections and whatnot. It works in its current state, but it's ugly – I found way more cross-module quirks than I was expecting. I wanted to make it as easy as possible for consumers and allow to use either version (with or without templates) from their SPM manifests.

Opinions, corrections, and alternative ideas are encouraged and most welcome!

---

How to use:

* If you want to use the core Tokenizers functionality, without the chat templates:

Add the following dependency to your target, as usual (except `Tokenizers` is now a product, no need to use the full `Transformers` lib). This applies to projects such as [WhisperKit](https://github.com/argmaxinc/WhisperKit).

```
            dependencies: [
                .product(name: "Tokenizers", package: "swift-transformers"),
            ],
```

* To opt-in to using chat-templates:

This applies to projects such as [mlx-swift-examples](https://github.com/ml-explore/mlx-swift-examples).

```diff
            dependencies: [
                .product(name: "Tokenizers", package: "swift-transformers"),
+               .product(name: "TokenizersTemplates", package: "swift-transformers"),
            ],
```

That's it, you don't need to import `TokenizersTemplates` or do anything other than just declaring it as a dependency.

---

Known issues:
- I haven't looked at tests yet, they probably won't compile.
- I know there are conflicts because of last night's tools PR. Let's agree on the general direction before addressing them.

</details>


Sources/Tokenizers/Tokenizer.swift
```diff
@@ -161,6 +160,20 @@ public protocol Tokenizer {
...
+
+    public func applyChatTemplate(messages: [[String: String]], chatTemplate: ChatTemplateArgument) throws -> [Int] {
+        try applyChatTemplate(messages: messages, chatTemplate: chatTemplate, addGenerationPrompt: true, truncation: false, maxLength: nil, tools: nil)
+    }
+
+    public func applyChatTemplate(messages: [[String: String]], chatTemplate: String) throws -> [Int] {
+        try applyChatTemplate(messages: messages, chatTemplate: .literal(chatTemplate), addGenerationPrompt: true, truncation: false, maxLength: nil, tools: nil)
+    }
+}
+
```
pcuenca, line 176: See comment in [`TokenizersTemplates`](https://github.com/huggingface/swift-transformers/pull/168/files#diff-d09b4e8a4286c2839f012b3e4d37d16468e70380679bb78808e0033210091b80R18)

Sources/TokenizersTemplates/TokenizersTemplates.swift
```diff
@@ -0,0 +1,115 @@
...
+    "unk_token",
+    "sep_token",
+    "pad_token",
+    "cls_token",
+    "mask_token",
+    "additional_special_tokens"
+]
+
+open class PreTrainedTokenizerWithTemplates : PreTrainedTokenizer {
+    // I don't know why these need to be here. They are implemented in the protocol, **and** in the superclass.
```
pcuenca, line 18: Yes, if these overrides don't exist, the linker can't find the implementations.

Sources/TokenizersWrapper/TokenizersWrapper.swift
```diff
@@ -0,0 +1,83 @@
+import Foundation
+import Hub
+
+@_exported import TokenizersCore
```
pcuenca, line 4: This is a good chunk of the magic. The new `Tokenizers` implementation is just this wrapper file, which exposes the imported `TokenizersCore` types.

Sources/TokenizersWrapper/TokenizersWrapper.swift
```diff
@@ -0,0 +1,83 @@
...
+
+#if canImport(TokenizersTemplates)
+import TokenizersTemplates
+public typealias PreTrainedTokenizer = PreTrainedTokenizerWithTemplates
+#endif
```
pcuenca, line 9: So if `TokenizersTemplates` is available (because users have declared it as a dependency), we override the definition so the factory below uses the subclass.

Sources/TokenizersWrapper/TokenizersWrapper.swift
```diff
@@ -0,0 +1,83 @@
...
+        let config = LanguageModelConfigurationFromHub(modelFolder: modelFolder, hubApi: hubApi)
+        guard let tokenizerConfig = try await config.tokenizerConfig else { throw TokenizerError.missingConfig }
+        let tokenizerData = try await config.tokenizerData
+
+        return try PreTrainedTokenizer(tokenizerConfig: tokenizerConfig, tokenizerData: tokenizerData)
+    }
+}
+
+// See https://github.com/xenova/transformers.js/blob/1a9964fb09b8f54fcbeac46dc6aae8d76795809d/src/tokenizers.js#L3203 for these exceptions
+class LlamaPreTrainedTokenizer: PreTrainedTokenizer {
```
pcuenca, line 65: This could be moved back to `TokenizersCore`, but then we'd need a `typealias` here as well (and a subclass).

pcuenca:
cc @greenrazer @FL33TW00D @Vaibhavs10 for opinions and feedback.

greenrazer:
I like this solution. however, I cannot get the `canImport` approach in `TokenizersWrapper` to work for the tests due to how Swift Package Manager handles module dependencies and conditional imports within the same package. All the tests work great with almost no modification (besides changing `Tokenizers` to `TokenizersCore`) except for the `ChatTemplatesTests.swift`.

I've tried a few different options:

- The best option: Just adding `TokenizersTemplates` as a dependency to the test, similar to how you would for an external package. This doesn’t work because it doesn’t trigger recompilation for the test.  
- Adding `swiftSettings: [.define("USE_TEMPLATES")]` to the test and then adding another build condition to `TokenizersWrapper`. Same issue as above—it doesn’t trigger recompilation for the test.  
- Creating two targets pointing to the same `TokenizersWrapper` source file, with one including `USE_TEMPLATES`. This doesn’t work because you can't have two targets referencing the same source file.  
- Adding `.target(name: "TokenizersTemplates", condition: .when(platforms: nil))` as a dependency to hopefully run only during testing. This doesn’t work in newer Swift versions.  
- Adding `swiftSettings: [.define("ENABLE_TEMPLATES", .when(configuration: .debug))]` to the `Tokenizers` target. This almost works, but `TokenizersTemplates` must then be a dependency of `Tokenizers`, which defeats the whole purpose.  

Since users must explicitly import `TokenizersTemplates` anyway, I think the best approach is to create two wrappers and two targets. That way, users only need to import either `TokenizersTemplates` or `Tokenizers`, depending on their needs.

Tests/TokenizersTests/TrieTests.swift
```diff
@@ -8,6 +8,8 @@
 @testable import Tokenizers
 import XCTest
 
+@testable import Tokenizers
+
```
pcuenca:
```suggestion
```

pcuenca:
@codex review

mattt:
@pcuenca Do you think we still need this after https://github.com/huggingface/swift-transformers/pull/240? If so, I'd be happy to give this another try in a new PR.

pcuenca:
> @pcuenca Do you think we still need this after #240? If so, I'd be happy to give this another try in a new PR.

You are right, I don't think we do. Can you please confirm @ZachNagengast? Jinja will now be coming from Hugging Face, same as swift-transformers.

pcuenca:
Closing this in favor of the much simpler #269. Will revisit if needed.

PR #169, opened 2025-01-30
Make `tools` and `additionalContext` optional

DePasqualeOrg:
I'm finishing up the [tools example](https://github.com/ml-explore/mlx-swift-examples/pull/174) in mlx-swift-examples, and it occurred to me that `tools` and `additionalContext` should probably be optional. Otherwise you might want to pass in an empty array, and that could affect how the template renders the prompt if it only checks if `tools` is defined, and not if `tools` is empty. What do you think, @pcuenca?


pcuenca (approved):
Yes, makes sense.

pcuenca:
I suppose we could also handle the empty array, but it's ok.

DePasqualeOrg:
Thank you! I'll be able to submit my tool use example in mlx-swift-examples once there's a new release tag for swift-transformers.

PR #170, opened 2025-01-31
Add the canonical cached token location

pcuenca:
I think ~/.huggingface is deprecated.


pcuenca:
cc @finnvoor in case of potential side effects. Sorry I didn't notice this when you submitted your PR.

PR #171, opened 2025-01-31
Public tokenizer errors, hasChatTemplate

pcuenca:
See https://github.com/ml-explore/mlx-swift-examples/issues/181#issuecomment-2607663177, https://github.com/ml-explore/mlx-swift-examples/issues/150

In addition to making the tokenizer errors public, this PR adds a new `hasChatTemplate` property. We could also create an accessor for the chat template itself if needed, but we still need to [support array values](https://github.com/huggingface/swift-transformers/blob/47d6c656dbdd7ec66d99f6110b8ecabf7c3a6f0f/Sources/Tokenizers/Tokenizer.swift#L459) that are in the process of being deprecated anyway.

Would these changes suffice for mlx-swift, @davidkoski? Do you think we'd need something like Python's `tokenize=False` argument, like @awni mentioned?

### DIFF
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -12,7 +12,7 @@ import Jinja
 public typealias Message = [String: Any]
 public typealias ToolSpec = [String: Any]
 
-enum TokenizerError: Error {
+public enum TokenizerError: Error {
     case missingConfig
     case missingTokenizerClassInConfig
     case unsupportedTokenizer(String)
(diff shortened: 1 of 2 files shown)


davidkoski:
Would `tokenize=False` be a method that just applied the template and returned the string?  The only thing I can think we need  it for today is if somebody wanted to log it for debug purposes -- perhaps the `llmtool` would call it and log it.

Right now the only problem I know of is the caller not knowing if they can safely call the applyChatTemplate method itself, so public error and `hasChatTemplate` seems sufficient.

pcuenca:
Yes, I think there are interesting use cases for `tokenize=False`, I've seen people decoding the ids as a workaround.

I'll merge this then and we can incorporate `tokenize=False` later. Thank you!

PR #173, opened 2025-02-09
Add detail progress for HubFileDownloader snapshot

everettjf:
When downloading, we usually want to get more detailed progress. For example, the download progress of each file and it's name.

The change is compatible with the original API.

I am a little unsure about the naming of the method. Feel free to comment.


FL33TW00D:
Thank you for your contribution! 

I think this will be superseded by #175?

PR #174, opened 2025-02-11
chore: experiment with swift-format for PR

FL33TW00D:
`swift-format` seems quite heavy handed.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,0 +1,87 @@
+--swiftversion 5.9
+--acronyms ID,URL,UUID
+--allman false
+--anonymousforeach convert
+--assetliterals visual-width
+--asynccapturing 
+--beforemarks 
+--binarygrouping 4,8
+--categorymark "MARK: %c"
+--classthreshold 0
+--closingparen balanced
+--closurevoid remove
+--commas always
+--conflictmarkers reject
+--decimalgrouping ignore
+--elseposition same-line
+--emptybraces spaced
+--enumthreshold 0
+--exponentcase lowercase
+--exponentgrouping disabled
+--extensionacl on-extension
+--extensionlength 0
+--extensionmark "MARK: - %t + %c"
+--fractiongrouping disabled
+--fragment false
+--funcattributes preserve
+--generictypes 
+--groupedextension "MARK: %c"
+--guardelse auto
+--header ignore
+--hexgrouping 4,8
+--hexliteralcase uppercase
+--ifdef no-indent
+--importgrouping alpha
+--indent 4
+--indentcase false
+--indentstrings false
+--lifecycle 
+--lineaftermarks true
+--linebreaks lf
+--markcategories true
+--markextensions always
+--marktypes always
+--maxwidth none
+--modifierorder 
+--nevertrailing 
+--nospaceoperators 
+--nowrapoperators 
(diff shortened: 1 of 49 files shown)


DePasqualeOrg:
@FL33TW00D, maybe try a line length of 120, which might result in fewer changes and looks better to my eye. If it's making too many changes that are not to your taste, it might be because it's applying default settings that you can override by specifying rules.

FL33TW00D:
@DePasqualeOrg if you have any suggestions on config to try and minimize changes that would be great! The main one for me is how to disable adding `public` everywhere, Apple that is not a formatter.

DePasqualeOrg:
It looks like the `public` additions are happening on extensions, and I don't see an option to avoid that in the [docs](https://github.com/swiftlang/swift-format/blob/main/Documentation/Configuration.md). I guess it's for clarity on things that are already public.

I don't care too much about specific rules, as long as they don't make the code harder to read. I find the 100-character line length makes the code less readable, and I have my configuration files set to 120 for that reason.

FL33TW00D:
@DePasqualeOrg Updated to 120 line length.

This can be tested locally with `act pull_request -W .github/workflows/format.yml -P macos-latest=-self-hosted`

shavit:
If finalized, consider adding a `.git-blame-ignore-revs` file with this commit.

DePasqualeOrg:
This issue has been in discussion for months, and the indecision continues to create problems for me as I try to contribute to this package. I would really appreciate it if you could decide on a solution and merge it.

FL33TW00D:
Sorry for the delay @DePasqualeOrg 

@pcuenca and myself will make sure some formatting solution is in by the end of the week.

FL33TW00D:
> If finalized, consider adding a `.git-blame-ignore-revs` file with this commit.

Good suggestion thank you!

FL33TW00D:
Overall I think we've decided to go with a gated CI approach using `swift-format` from [Nick Lockwood](https://github.com/nicklockwood/SwiftFormat). I think this approach is superior to a pre-commit hook approach as it ensures consistency, and we can reduce the downside of back and forth by having the CI comment on the PR with instructions to fix the formatting issues.

The suggestion from @shavit is solid, and we can add a note about the `.git-blame-ignore-revs` in the README.

If @pcuenca agrees then hopefully we can go for it!

pcuenca (commented):
Yes, CI formatting is enough imo, we can also provide a local script to run the steps locally if we need to.

Can we have all the changes applied here so we can see what the result looks like?

pcuenca (approved):
We already found:
- A compilation problem because one of the rules (using keypaths) should not have been applied.
- Blank space changes in literal Strings 😱 

I hope this is useful to others, but it sure is scarier than it should (and a time sink already).

DePasqualeOrg:
Are you referring to the multi-line string in `testSuccessfulDownload`? That was pretty easy to resolve. Multi-line strings can be tricky, but now we can be sure that they will be correct.

FL33TW00D:
> Are you referring to the multi-line string in `testSuccessfulDownload`? That was pretty easy to resolve. Multi-line strings can be tricky, but now we can be sure that they will be correct.

https://github.com/nicklockwood/SwiftFormat/issues/1993

PR #175, opened 2025-02-11
Add resumable download support with tests

ardaatahan:
This PR aims to bring over huggingface_hub Python library's resumable downloads logic to swift-transformers:

- Implements chunked transfer and retry logic for more reliable downloads
- Adds support for resumable downloads through range headers
- Adds configurable parameters for timeout, retries, and chunk size
- Improves error handling and size verification for downloads
- Uses temporary files during download to prevent corruption of existing files
- Adds new tests to verify functionality of resumable downloads
- Fixes testDeepSeekQwenChatTemplate function which was previously validating an outdated chat template output. The test now includes the correct tokens, referenced from the tokenizer repo

### DIFF
--- a/Sources/Hub/Downloader.swift
+++ b/Sources/Hub/Downloader.swift
@@ -71,11 +71,134 @@ class Downloader: NSObject, ObservableObject {
                 }
             }
             var request = URLRequest(url: url)
+            
+            // Use headers from argument else create an empty header dictionary
+            var requestHeaders = headers ?? [:]
+            
+            // Populate header auth and range fields
             if let authToken = authToken {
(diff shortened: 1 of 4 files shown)


pcuenca:
cc @greenrazer, would you like to give it a go at testing / reviewing this?

FL33TW00D:
Just tested this out, looks great thank you for the contribution!

DePasqualeOrg:
@FL33TW00D, just a suggestion, but maybe squash the commits before merging next time, so that the repo's commit history doesn't get cluttered. I think you can also set this as a setting on the repo.

pcuenca:
Changed! Sorry, it used to be just me merging PRs :)

PR #177, opened 2025-02-14
Add offline mode with tests

ardaatahan:
This PR aims to bring offline mode logic to swift-transformers:

- Adds useOfflineMode flag to snapshot function to give developers explicit control over offline behavior without relying on automatic network detection 
- Adds NetworkMonitor class to detect network connectivity state (connected/expensive/constrained), which is useful to decide whether to use offline mode when the developers have no explicit preference 
- Implements offline file validation by checking local files, metadata integrity, and LFS file hashes
- Adds new OfflineModeError cases and improves error handling
- Adds FileManager extension to support recursive file enumeration
- Reorganizes code by moving metadata helper functions to HubApi level and simplifying HubFileDownloader, which makes them easier to use for offline mode file verification

These changes enable the library to work reliably without network access and help reduce bandwidth usage on mobile devices with limited connectivity or cellular data plans. This is particularly important for ML models which can be quite large, ensuring users don't unexpectedly consume large amounts of mobile data when models are already present locally.

### DIFF
--- a/Sources/Hub/HubApi.swift
+++ b/Sources/Hub/HubApi.swift
@@ -161,10 +161,12 @@ public extension HubApi {
 public extension HubApi {
     enum EnvironmentError: LocalizedError {
         case invalidMetadataError(String)
-        
+        case offlineModeError(String)
+                    
         public var errorDescription: String? {
             switch self {
-            case .invalidMetadataError(let message):
+            case .invalidMetadataError(let message),
+                 .offlineModeError(let message):
                 return message
             }
         }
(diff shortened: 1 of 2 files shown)


FL33TW00D:
Seems `testDeepSeekQwenChatTemplate` is failing here

ardaatahan:
It's because the test case in main is validating an outdated chat template output. I updated that test case to validate the correct output tokens in this pr: https://github.com/huggingface/swift-transformers/pull/175

ZachNagengast:
Yea the template itself changed on the source repo - see this commit https://huggingface.co/deepseek-ai/DeepSeek-R1-Distill-Qwen-7B/commit/6602cadec947dbb53e64f3d8d6425320b2197247 #175 has the fix but  could also be part of its own PR, since main tests would currently fail if ran again due the them changing the template in the model repo

FL33TW00D:
Please rebase when you're ready now that #175 is merged :)

ardaatahan:
@FL33TW00D thanks! I just merged the recent changes

FL33TW00D:
Happy to merge this if we can confirm that #180 is solved!

PR #179, opened 2025-02-18
feat: Update swift-argument-parser to version 1.5.0 or higher

matiasvillaverde:
Update swift-argument-parser to version 1.5.0 or higher

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -12,7 +12,7 @@ let package = Package(
         .executable(name: "hub-cli", targets: ["HubCLI"]),
     ],
     dependencies: [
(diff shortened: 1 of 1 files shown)


mattt:
Hi @matiasvillaverde. Thanks for opening this PR. Your change was indeed a better way to pull in `swift-argument-parser.

I'm happy to report that we've gone ahead and removed that dependency entirely with https://github.com/huggingface/swift-transformers/pull/235 (motivated in part because of the challenges you highlighted in this PR). So we shouldn't have this problem going forward 😄

PR #183, opened 2025-02-25
Match token value with array and tuple

### LINKED ISSUE #182: fatalError("Missing `sep` processor configuration" when using "facebook/bart-large-mnli"
When loading tokenizer using AutoTokenizer it's crashing on this line.

https://github.com/huggingface/swift-transformers/blob/1d26ce0b7706b2492421079849e244f626abd944/Sources/Tokenizers/PostProcessor.swift#L95

shavit:
* Add match cases for token value
  * Add BartTokenizer to known tokenizers

/Close #182

### DIFF
--- a/Tests/HubTests/HubTests.swift
+++ b/Tests/HubTests/HubTests.swift
@@ -118,4 +118,14 @@ class HubTests: XCTestCase {
         let vocab_dict = config.dictionary["vocab"] as! [String: Int]
         XCTAssertNotEqual(vocab_dict.count, 2)
     }
+
+    func testConfigTokenValueDifferentOrder() {
+        let data: Data! = "{\"sep\": [\"</s>\", 2], \"cls\": [0, \"<s>\"]}".data(using: .utf8)
+        let dict = try! JSONSerialization.jsonObject(with: data, options: []) as! [NSString: Any]
+        let config = Config(dict)
+        XCTAssertEqual(config.sep!.tokenValue!.0, 2)
+        XCTAssertEqual(config.sep!.tokenValue!.1, "</s>")
+        XCTAssertEqual(config.cls!.tokenValue!.0, 0)
+        XCTAssertEqual(config.cls!.tokenValue!.1, "<s>")
+    }
 }
--- a/Sources/Hub/Hub.swift
+++ b/Sources/Hub/Hub.swift
@@ -100,7 +100,14 @@ public struct Config {
     }
     
     /// Tuple of token identifier and string value
-    public var tokenValue: (UInt, String)? { value as? (UInt, String) }
+    public var tokenValue: (UInt, String)? { 
+        switch value {
+        case let (i, t) as (UInt, String): return (i, t)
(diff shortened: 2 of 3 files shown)


greenrazer:
Thank you for the contribution @shavit! I tested it and `facebook/bart-large-mnli` (#182) is now working with the AutoTokenizer.

However this is a curious issue, because when pulling from a URL `AutoTokenzier.from` calls:

- `LanguageModelConfigurationFromHub.init`
- Then `LanguageModelConfigurationFromHub.loadConfig`
- Then `calls this code from HubAPI.swift`:

```(swift)
func configuration(fileURL: URL) throws -> Config {
    let parsed = try JSONSerialization.jsonObject(with: data, options: [])
    guard let dictionary = parsed as? [NSString: Any] else { throw Hub.HubClientError.parse }
    return Config(dictionary)
}
```

I could be missing something, but since JSON doesn't support tuples, wouldn't we expect the value entries in the config to be arrays instead of tuples? I'm a bit confused why this worked previously. @pcuenca @FL33TW00D

shavit:
Right, I agree. The config likely decodes to a dictionary with any byte values, and the token value casts in a dynamic lookup. 

I'll do a few more checks.

pcuenca:
I think you are right @greenrazer! I don't think this could ever work with actual tokenizer files. `sep` and `cls` would be decoded as arrays, and I don't think we can cast them to tuples.  The correct way to do it is the current one proposed in this PR.

shavit:
True, but the JSON array was decoded into `Any?`, not a tuple. The value is then read from a new `Config` and matched with the correct type. 
https://github.com/huggingface/swift-transformers/blob/main/Sources/Hub/Hub.swift#L83

The problem comes from the token order, which is not index-first.
```
{
  "type": "RobertaProcessing",
  "sep": [
    "</s>",
    2
  ],
  "cls": [
    "<s>",
    0
  ],
}
```

Tests/HubTests/HubTests.swift
```diff
@@ -118,4 +118,14 @@ class HubTests: XCTestCase {
         let vocab_dict = config.dictionary["vocab"] as! [String: Int]
         XCTAssertNotEqual(vocab_dict.count, 2)
     }
+
+    func testConfigTokenValueDifferentOrder() {
```
pcuenca, line 122: Thanks for the test! Could we please add another one that loads a tokenizer such as `facebook/bart-large-mnli`? We don't need anything fancy: just show that the tokenizer loads and that it can encode a string.

pcuenca (commented):
Nice! Could we add a simple test with a

jkrukowski:
Added the discussed parsing in this PR https://github.com/huggingface/swift-transformers/pull/188
If the parsing to tuple is not needed I could remove it, lmk what you think @pcuenca and @greenrazer

FL33TW00D:
@jkrukowski added this to his PR #188 so I will close this @shavit but thank you for your contribution!

PR #184, opened 2025-02-26
Prefer chat_template.json for chat template

DePasqualeOrg:
https://github.com/ml-explore/mlx-swift-examples/pull/197#issuecomment-2683459124

### DIFF
--- a/Sources/Hub/Hub.swift
+++ b/Sources/Hub/Hub.swift
@@ -183,45 +183,33 @@ public class LanguageModelConfigurationFromHub {
 
         return try await loadConfig(modelFolder: downloadedModelFolder, hubApi: hubApi)
     }
-    
+
     func loadConfig(
         modelFolder: URL,
         hubApi: HubApi = .shared
     ) async throws -> Configurations {
-        // Note tokenizerConfig may be nil (does not exist in all models)
+        // Load required configurations
         let modelConfig = try hubApi.configuration(fileURL: modelFolder.appending(path: "config.json"))
-        // First try to get the tokenizer_config.json
-        let tokenizerConfig = try? hubApi.configuration(fileURL: modelFolder.appending(path: "tokenizer_config.json"))
-        // Check for chat_template.json, which contains the preferred chat template for vision language models
-        if let chatTemplateConfig = try? hubApi.configuration(fileURL: modelFolder.appending(path: "chat_template.json")) {
-            // If chat_template.json exists and contains a chat_template field, use it to override the tokenizer_config
-            if let chatTemplate = chatTemplateConfig.chatTemplate?.stringValue {
-                var updatedConfig: Config
-                if var configDict = tokenizerConfig?.dictionary {
-                    // Override the chat template in the existing tokenizer config
-                    configDict["chat_template"] = chatTemplate
-                    updatedConfig = Config(configDict)
-                } else {
-                    // Create a new config with just the chat template
-                    updatedConfig = Config(["chat_template": chatTemplate])
-                }
-                let tokenizerVocab = try hubApi.configuration(fileURL: modelFolder.appending(path: "tokenizer.json"))
-                let configs = Configurations(
-                    modelConfig: modelConfig,
-                    tokenizerConfig: updatedConfig,
-                    tokenizerData: tokenizerVocab
-                )
-                return configs
+        let tokenizerData = try hubApi.configuration(fileURL: modelFolder.appending(path: "tokenizer.json"))
+        // Load tokenizer config
+        var tokenizerConfig = try? hubApi.configuration(fileURL: modelFolder.appending(path: "tokenizer_config.json"))
+        // Check for chat template and merge if available
+        if let chatTemplateConfig = try? hubApi.configuration(fileURL: modelFolder.appending(path: "chat_template.json")),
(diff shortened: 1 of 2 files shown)


FL33TW00D:
Looks good to me @pcuenca

Sources/Hub/Hub.swift
```diff
@@ -190,9 +190,32 @@ public class LanguageModelConfigurationFromHub {
...
+                    // Create a new config with just the chat template
+                    updatedConfig = Config(["chat_template": chatTemplate])
+                }
+                let tokenizerVocab = try hubApi.configuration(fileURL: modelFolder.appending(path: "tokenizer.json"))
+                let configs = Configurations(
+                    modelConfig: modelConfig,
+                    tokenizerConfig: updatedConfig,
+                    tokenizerData: tokenizerVocab
+                )
+                return configs
```
pcuenca, line 214: Not a fan of repeating this code block and the return here, inside the nested ifs.

I'd recommend we write a helper function to potentially update the chat template, such as:

```swift
    func updatedTokenizerConfig(tokenizerConfig: Config?, chatTemplateConfig: Config?) -> Config? {
        guard
            let chatTemplateConfig = chatTemplateConfig,
            let overrideChatTemplate = chatTemplateConfig.chatTemplate?.stringValue else {
            return tokenizerConfig
        }

        var configDict = tokenizerConfig?.dictionary ?? [:]
        configDict["chat_template"] = overrideChatTemplate
        return Config(configDict)
    }
```

And then we can just use this before the return:

```swift
        let configs = Configurations(
            modelConfig: modelConfig,
            tokenizerConfig: updatedTokenizerConfig(tokenizerConfig: tokenizerConfig, chatTemplateConfig: chatTemplateConfig),
            tokenizerData: tokenizerVocab
        )
```

Tests/TokenizersTests/ChatTemplateTests.swift
```diff
@@ -178,6 +178,43 @@ What is the weather in Paris today?<|im_end|>
...
+        let expectedOutput = """
+<|im_start|>system
+You are a helpful assistant.<|im_end|>
+<|im_start|>user
+What's in this image?<|vision_start|><|image_pad|><|vision_end|><|im_end|>
+<|im_start|>assistant
+
+"""
+        XCTAssertTrue(qwen2VLEncoded == qwen2_5VLEncoded)
+        XCTAssertTrue(qwen2VLDecoded == qwen2_5VLDecoded && qwen2_5VLDecoded == expectedOutput)
```
pcuenca, line 215:
```suggestion
        XCTAssertEqual(qwen2VLEncoded, qwen2_5VLEncoded, "Encoded sequences should be equal")
        XCTAssertEqual(qwen2VLDecoded, qwen2_5VLDecoded, "Decoded sequences should be equal")
        XCTAssertEqual(qwen2_5VLDecoded, expectedOutput, "Decoded should match expected")
```

nit, should provide better error messages maybe

Sources/Hub/Hub.swift
```diff
@@ -190,9 +190,32 @@ public class LanguageModelConfigurationFromHub {
     ) async throws -> Configurations {
         // Note tokenizerConfig may be nil (does not exist in all models)
         let modelConfig = try hubApi.configuration(fileURL: modelFolder.appending(path: "config.json"))
+        // First try to get the tokenizer_config.json
         let tokenizerConfig = try? hubApi.configuration(fileURL: modelFolder.appending(path: "tokenizer_config.json"))
+        // Check for chat_template.json, which contains the preferred chat template for vision language models
+        if let chatTemplateConfig = try? hubApi.configuration(fileURL: modelFolder.appending(path: "chat_template.json")) {
```
pcuenca, line 196: Technically, this is not the same algorithm used in transformers. IIRC, if we instantiate a tokenizer from a repo where the tokenizer has a chat template and a different `chat_template.json`, the template from the tokenizer will still be used. However, if we instantiate a `processor`, then the `chat_template.json` will be used.

I'm willing to diverge from this behaviour, given that:
- Chat template divergence should not be expected (it's possibly a mistake if both templates differ)
- Processor and tokenizer templates should be synced at some point.
- There is no `processor` abstraction in swift-transformers.

cc @Rocketknight1 in case I'm missing some weird edge case.

Sources/Hub/Hub.swift
```diff
@@ -190,9 +190,32 @@ public class LanguageModelConfigurationFromHub {
...
+            // If chat_template.json exists and contains a chat_template field, use it to override the tokenizer_config
+            if let chatTemplate = chatTemplateConfig.chatTemplate?.stringValue {
```
pcuenca, line 198: Technically, this could potentially be an array too. But this is discouraged.

pcuenca (commented):
Thanks for another great contribution @DePasqualeOrg!

There are a couple of subtleties about chat templates that I'm willing to ignore (unless I'm missing something important), and some minor code changes that I think could make the code a bit cleaner – although this could admittedly be subjective.

DePasqualeOrg:
Thanks for your suggestions, @pcuenca. I made some changes based on your input.

pcuenca (approved):

DePasqualeOrg:
We'll need a new version tag in this repo to pick this up in mlx-swift-examples so that we can make Qwen 2.5 VL work with the correct chat template.

pcuenca:
Added

PR #196, opened 2025-04-07
Use injected HubApi for downloads instead of shared

petrukha-ivan:
The current implementation of `HubFileDownloader` uses a shared instance of `HubApi`, which can lead to authorization errors when accessing private repositories due to a missing token. This PR updates the implementation to correctly respect and use the provided `HubApi` instance, ensuring proper authentication.


FL33TW00D:
@petrukha-ivan great work! thank you for your contribution.

PR #198, opened 2025-04-25
Add incomplete download support

ardaatahan:
This PR:

- Expands on #187 which downloads incomplete files directly into the target directory, which could lead to runtime errors while trying to read those files in the case a download did not fully complete. This is resolved by downloading incomplete files first to the cache directory then moving them to their final locations. This increases swift-transformers' download logic parity with the huggingface_hub library as laid out [here](https://github.com/huggingface/swift-transformers/pull/187#issuecomment)
- Also adds download cancellation support. Previously, if you were downloading a file inside a task and cancelled the task, the download would continue to run even if the task was canceled. This is done via adding `withTaskCancellationHandler` inside `HubFileDownloader`'s `Download`  method, which would immediately invalidate the url session and updates progress as cancelled.
- Renames urlSession to session in `Downloader` class to resolve naming conflict with URLSessionDownloadDelegate methods with the same name.
- Creates target directory structure lazily (as opposed to eagerly before) when moving the incomplete files to their final locations.
- Adds/improves test cases to verify new functionality.

### DIFF
--- a/Sources/Hub/Downloader.swift
+++ b/Sources/Hub/Downloader.swift
@@ -11,7 +11,6 @@ import Foundation
 
 class Downloader: NSObject, ObservableObject {
     private(set) var destination: URL
-    private(set) var sourceURL: URL
 
     private let chunkSize = 10 * 1024 * 1024 // 10MB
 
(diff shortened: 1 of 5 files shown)


ardaatahan:
@pcuenca, @FL33TW00D FYI - I changed the `session` member variable (renamed from `urlSession` due to naming conflict) in `Downloader` to internal from private. This is because the `DownloaderTests` file extends `Downloader` with a test only `interruptDownload` method that directly accesses the `session` to simulate network interruptions at precise progress checkpoints (50% and 75%). This allows verification of the resumable download functionality I added with #155. Since this access remains module internal and doesn't expand the public API surface, it still maintains proper encapsulation.

FL33TW00D:
Looks really good, nice work @ardaatahan

PR #200, opened 2025-05-21
Fix formatting

DePasqualeOrg:
A bunch of extra whitespace on empty lines was added in the last commit. This fixes the issue throughout the repo by using `--trimwhitespace always` instead of `--trimwhitespace nonblank-lines`.

PR #202, opened 2025-05-23
Add repo revision option

petrukha-ivan:
### What 
Added `revision` parameter to specify a branch, tag, or commit when downloading a model.

### DIFF
--- a/Sources/Hub/HubApi.swift
+++ b/Sources/Hub/HubApi.swift
@@ -335,6 +335,7 @@ public extension HubApi {
     struct HubFileDownloader {
         let hub: HubApi
         let repo: Repo
+        let revision: String
         let repoDestination: URL
         let repoMetadataDestination: URL
         let relativeFilename: String
@@ -349,7 +350,7 @@ public extension HubApi {
                 url = url.appending(component: repo.type.rawValue)
             }
             url = url.appending(path: repo.id)
-            url = url.appending(path: "resolve/main") // TODO: revisions
+            url = url.appending(path: "resolve/\(revision)")
             url = url.appending(path: relativeFilename)
             return url
         }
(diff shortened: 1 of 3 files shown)


Sources/Hub/HubApi.swift
```diff
@@ -349,7 +350,7 @@ public extension HubApi {
                 url = url.appending(component: repo.type.rawValue)
             }
             url = url.appending(path: repo.id)
-            url = url.appending(path: "resolve/main") // TODO: revisions
+            url = url.appending(path: "resolve/\(revision)")
```
FL33TW00D, line 353: Maybe best not to do direct string interpolation?
Also needs a null check I think

petrukha-ivan: String interpolation is the preferred method for this in Swift because it's faster and benefits from compile-time optimizations. We also don't need a null check here since revision is not marked as `Optional` and always has a value. However, I can add something like `assert(revision.isEmpty == false, "Revision cannot be empty")` if you'd prefer 🙂

FL33TW00D:
Just tested:
```
~/Code/swift-transformers test-revision ❯ .build/release/hub-cli download jinaai/jina-embeddings-v3 --revision BINGO 
Error: Authentication required. Please provide a valid Hugging Face token.
~/Code/swift-transformers test-revision ❯ .build/release/hub-cli download jinaai/jina-embeddings-v3 --revision main 
40.00%
~/Code/swift-transformers test-revision ❯ .build/release/hub-cli download jinaai/jina-embeddings-v3 --revision INVALID$%£@ 
Error: Authentication required. Please provide a valid Hugging Face token.
```
Works fine with valid revision, incorrect error message if not. Would add a test with janky characters too, can never be too sure!

petrukha-ivan:
@FL33TW00D Hi! You're right 👍 I've fixed the error type to throw the correct error in case of invalid revision, and also added a few unit tests to test both positive and negative cases.

pcuenca (approved):
Looks good to me, thanks for the great contribution and for iterating 🙌 !

I think we should maybe merge #200 first, just in case. Would you agree, @FL33TW00D?

FL33TW00D:
Still seeing an issue here @petrukha-ivan, try the following:
`.build/release/hub-cli download jinaai/jina-embeddings-v3 --revision main`
`.build/release/hub-cli download jinaai/jina-embeddings-v3 --revision refactor-task-type`

petrukha-ivan:
@FL33TW00D Ah, good catch — it's fixed now 👍

pcuenca:
One last thing: `getFileMetadata` is still using `main`, I think. Can we pass revision through to it?

FL33TW00D:
@petrukha-ivan Nothing you've done, we are getting 429'd by the Hub

FL33TW00D:
@petrukha-ivan Appreciate the back and forth here! Thanks for the contribution 🤗

PR #206, opened 2025-06-02
Add some backwards-compatible getters

pcuenca:
Unfortunately, we can't provide a `dictionary` var because the symbol has been replaced with `dictionary()` now :(

Even so, I think this could be useful for many downstream users, I don't think the use of `dictionary` is too frequent.

What are your thoughts @FL33TW00D, @DePasqualeOrg?

### DIFF
--- a/Sources/Hub/Config.swift
+++ b/Sources/Hub/Config.swift
@@ -616,6 +616,24 @@ public struct Config: Hashable, Sendable,
     }
 }
 
+// Old style, deprecated getters
+extension Config {
+    @available(*, deprecated, message: "Use string() instead")
+    public var stringValue: String? { string() }
+
+    @available(*, deprecated, message: "Use integer() instead")
+    public var intValue: Int? { integer() }
+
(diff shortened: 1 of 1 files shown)


FL33TW00D:
@pcuenca Looks really good! 

CI is only failing due to 429.

DePasqualeOrg:
I didn't have time to dig into this, so no notes. I'll test it in https://github.com/ml-explore/mlx-swift-examples/pull/322 once it's merged.

pcuenca:
> I didn't have time to dig into this

TL;DR: you will be able to keep using `stringValue`, but `dictionary` will still have to be changed to `dictionary()`, unfortunately.

DePasqualeOrg:
Great, https://github.com/ml-explore/mlx-swift-examples/pull/322 builds with no errors using the latest commit of swift-transformers. Would you like to create a new version tag? Then I can update the dependency in mlx-swift-examples.

PR #207, opened 2025-06-03
updating Jinja to 1.2.0 as it correctly supports Granite model template

smdesai:
Updating the version to correctly handle Granite model templates.


FL33TW00D:
Thanks!

PR #224, opened 2025-09-02
Invoke `swiftformat` via Swift Package Manager plugin 

mattt:
The [`format` workflow](https://github.com/huggingface/swift-transformers/blob/ed54f3c9589d6cdc6fc2bb810e4809b0735fee01/.github/workflows/format.yml) currently defines the current SwiftFormat version as an environment variable:

https://github.com/huggingface/swift-transformers/blob/ed54f3c9589d6cdc6fc2bb810e4809b0735fee01/.github/workflows/format.yml#L8-L9

As we see in #223, defining a constant here creates drift over time. 

This PR proposes to declare SwiftFormat as a package dependency and call it as a [Swift Package Manager plugin](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/plugins). This approach defines a single source of truth for managing the development dependency, and provide developers and CI a unified way to apply and lint project standards.

**Before**

```bash
brew install swiftformat # or curl && unzip && etc.
swiftformat --config .swiftformat .
```

**After**

```bash
swift package plugin --allow-writing-to-package-directory swiftformat --config .swiftformat .
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,6 +19,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMinor(from: "1.4.0")),
         .package(url: "https://github.com/johnmai-dev/Jinja", .upToNextMinor(from: "1.2.1")),
+        .package(url: "https://github.com/nicklockwood/SwiftFormat", from: "0.54.0"),
     ],
     targets: [
         .executableTarget(
--- a/Package.swift
+++ b/Package.swift
@@ -19,7 +19,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMinor(from: "1.4.0")),
         .package(url: "https://github.com/johnmai-dev/Jinja", .upToNextMinor(from: "1.2.1")),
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -19,6 +19,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser.git", .upToNextMinor(from: "1.4.0")),
         .package(url: "https://github.com/johnmai-dev/Jinja", .upToNextMinor(from: "1.2.1")),
+        .package(url: "https://github.com/nicklockwood/SwiftFormat", from: "0.57.0"),
```
pcuenca, line 22: So this would unconditionally download the dependency and compile it on the fly when `swift package plugin` is invoked?

Could download be potentially avoided hiding it behind an opt-in `traits` wall? (Just to understand it better, still testing that approach in its PR)

mattt: Swift Package Manager automatically downloads all declared dependencies whenever it resolves the dependency graph — so, anytime you run `swift build` and `swift package plugin`. My understanding is that it's built on-demand only when you invoke `swift package plugin`. Unfortunately, I don't believe traits would help us here.

mattt:
Closing in favor of https://github.com/huggingface/swift-transformers/pull/248

PR #229, opened 2025-09-10
Migrate from XCTest to Swift Testing

mattt:
This PR migrates most of our tests from XCTest to the new Swift Testing framework, which has better ergonomics and performance.

### DIFF
--- a/Tests/TokenizersTests/ChatTemplateTests.swift
+++ b/Tests/TokenizersTests/ChatTemplateTests.swift
@@ -309,7 +309,7 @@ struct ChatTemplateTests {
 
     @Test("Apply template error handling")
     func applyTemplateError() async throws {
-        let tokenizer = try await AutoTokenizer.from(pretrained: "google-bert/bert-base-uncased")
+        let tokenizer = try await AutoTokenizer.from(pretrained: "google-bert/bert-base-chinese")
         #expect(!tokenizer.hasChatTemplate)
 
         #expect(throws: TokenizerError.self) {


Tests/GenerationTests/TestUtils.swift
```diff
@@ -1,22 +1,7 @@
...
-        let rhsEvaluated = try expression2()
-        XCTAssertEqual(lhsEvaluated.count, rhsEvaluated.count, file: file, line: line)
-        for (lhs, rhs) in zip(lhsEvaluated, rhsEvaluated) {
-            XCTAssertEqual(lhs, rhs, accuracy: accuracy, file: file, line: line)
-        }
-    } catch {
-        XCTFail("Unexpected error: \(error)", file: file, line: line)
-    }
+/// Check if two floating-point arrays are equal within a given accuracy
+func isClose<T: FloatingPoint>(_ lhs: [T], _ rhs: [T], accuracy: T) -> Bool {
```
pcuenca, line 4: Nice. This could become a general utility function later, like Pytorch or MLX `allClose`.

pcuenca:
> Are parallel tokenizer tests still revealing race conditions?

Ah, I see they work now!

Tests/GenerationTests/TestUtils.swift
mattt: Yeah, that would be nice. Official guidance from the Swift team for replacing the XCTest assertion with accuracy is to pull in swift-numerics to use their implementation of [`isApproximatelyEqual`](https://github.com/apple/swift-numerics/blob/be0d7c65d1321e0cddee84acba10daf5f13ff795/Sources/RealModule/ApproximateEquality.swift#L72). Pulling that in as a separate dependency just for tests is overkill, but we could certainly vendor it.

mattt:
> > Are parallel tokenizer tests still revealing race conditions?
> 
> Ah, I see they work now!

Yes, that's the race conditions I was mentioning before. There's a more comprehensive change we can do to make that reentrancy safe, but for now it's much easier to ensure that we don't have more than one test calling `AutoTokenizer.from(pretrained:)` with the same model.

mattt:
Ooh, nice — after making everything parallel again, we're now down to 3 minutes 40 seconds 🚀

PR #234, opened 2025-09-11
Fix legacy behaviour

### LINKED ISSUE #96: Tokenizer behavior is different from Python transformers
Hi, I found the tokenizer behavior different from Python transformers when I use Phi-3 model.

swift-transformers
```swift
func testTokenizer() async throws {
    let tokenizer = try await AutoTokenizer.from(pretrained: "mlx-community/Phi-3-mini-4k-instruct-4bit-no-q-embed")
    let inputIds = tokenizer(" Hi")
    print(inputIds)
    // output: [1, 6324]
}
```

Python transformers
```python
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("mlx-community/Phi-3-mini-4k-instruct-4bit-no-q-embed")
input_ids = tokenizer.encode(" Hi")
print(input_ids)
# output: [1, 29871, 6324]
```

Python transformers prepends `29871` (`▁`) before `6324`. It seems to be done by the normalizer. I debugged this issue and found that the normalizer is ignored when `legacy` is `false` at https://github.com/huggingface/swift-transformers/blob/fc6543263e4caed9bf6107466d625cfae9357f08/Sources/Tokenizers/Tokenizer.swift#L341-L344.

pcuenca:
There was a case I had not considered 😢

Fixes #96

### DIFF
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -722,7 +722,7 @@ class LlamaPreTrainedTokenizer: PreTrainedTokenizer {
         try super.init(tokenizerConfig: tokenizerConfig, tokenizerData: updatedData, strict: strict)
     }
 
-    /// If `self.legacy` is set to `False`, a prefix token is added unless the first token is special.
+    /// If `isLegacy` is `False`, a prefix token is added unless the first token is special.
     /// https://github.com/huggingface/transformers/blob/e6dcf8abd6f65bb4b6dfc1831b20d9ba49ce00e2/src/transformers/models/t5/tokenization_t5.py#L374-L387
     override func tokenize(text: String) -> [String] {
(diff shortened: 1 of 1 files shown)


Sources/Tokenizers/Tokenizer.swift
```diff
@@ -721,4 +721,20 @@ class LlamaPreTrainedTokenizer: PreTrainedTokenizer {
         let updatedData = Config(configDictionary)
         try super.init(tokenizerConfig: tokenizerConfig, tokenizerData: updatedData, strict: strict)
     }
+
+    /// If `self.legacy` is set to `False`, a prefix token is added unless the first token is special.
```
pcuenca:
```suggestion
    /// If `isLegacy` is `False`, a prefix token is added unless the first token is special.
```

PR #239, opened 2025-09-16
Fix repo file update

petrukha-ivan:
# What
Currently, if you update any file in a repo after the initial download and then try to update local files, it fails with an error:

```
Error Domain=NSCocoaErrorDomain Code=516 "{file}.incomplete" couldn't be moved to "{destination}" because an item with the same name already exists.
```

The problem is in how the check for an existing local file is done:

```swift
fileExists(atPath: dstURL.path()) // false
fileExists(atPath: dstURL.path(percentEncoded: false)) // true
```

### DIFF
--- a/Sources/Hub/Downloader.swift
+++ b/Sources/Hub/Downloader.swift
@@ -347,7 +347,7 @@ extension Downloader: URLSessionDownloadDelegate {
 
 extension FileManager {
     func moveDownloadedFile(from srcURL: URL, to dstURL: URL) throws {
-        if fileExists(atPath: dstURL.path()) {
+        if fileExists(atPath: dstURL.path(percentEncoded: false)) {
             try removeItem(at: dstURL)
         }
 


pcuenca:
Interesting, we have a [couple of tests](https://github.com/huggingface/swift-transformers/blob/f6ca31803f3ad00a19c6b845d37d95694bd583a6/Tests/HubTests/HubApiTests.swift#L779) for this. Do you have an example filename that fails @petrukha-ivan? Maybe just a filename with a space?

petrukha-ivan:
The filename is just a regular `config.json`, but the destination directory is `~/Library/Application Support`, which causes the issue.

pcuenca:
Nice, thank you! Would you have time to take a look @FL33TW00D?

FL33TW00D:
This looks great! Thank you @petrukha-ivan

pcuenca:
Perhaps we should add a test for this case.

PR #189, opened 2025-03-07
WIP: Sendable Config

piotrkowalczuk:
This PR is a follow-up for [#172](https://github.com/huggingface/swift-transformers/issues/172) , it introduces a `Sendable` `Config`, built around `BinaryDistinctString`, `BinaryDistinctDictionary`, and `BinaryDistinctCharacter`.

## Motivation
- Ensure `Config` conforms to [`Sendable`](https://developer.apple.com/documentation/swift/sendable).
- Provide an explicit API instead of relying on `@dynamicMemberLookup`.
- Minimize changes to business logic unless strictly necessary.

## TODO
- [x] Ensure all existing tests pass.
- [ ] Implement JSON serialization for `Config`, particularly in chat template contexts.
- [ ] Add explicit tests for UTF-16 handling and protocol conformance.
- [ ] Review and address TODO comments in the code.
- [ ] Remove any unused code.

## Open Questions
- [ ] Should `BinaryDistinctString` and `BinaryDistinctCharacter` be adopted more widely across the package, replacing `String` and `NSString` where applicable?
  - In practice, `String` already contains all necessary information for binary-distinct comparisons.
  - The key difference lies in how the `==` operator affects serialization, deserialization, and hashing for uniqueness.
- [ ] Making tokenizers `Sendable` would require deep changes to the Tokenizers package. What other improvements should be incorporated next to the https://github.com/huggingface/swift-transformers/pull/168? 

## Follow-ups
- Improve the `Hub` package, particularly by making `Downloader` conform to `Sendable` so the entire target can be marked as Swift6 compatible. 
- Redesign `Tokenizers` package.

PR #245, opened 2025-09-19
Remove `UnitTests.xctestplan`

mattt:
This file was added in #189, perhaps unintentionally. `.xctestplan` files are Xcode-specific; Swift Package Manager has its own testing mechanism (`swift test`) that doesn't use these files. 

I've noticed that, when opening the package in Xcode, the existence of this file around has caused weird conflicts with the test plan Xcode was attempting to generate. For example, with #244, removing the `TestUtilsTests` target, which was defined in the test plan, caused errors in Xcode's test navigator.


FL33TW00D:
Nice!

pcuenca:
Thank you!

PR #246, opened 2025-09-19
Make consistent module resource locations

mattt:
The `Hub` module contains a `FallbackConfigs` directory containing fallback tokenizer configurations whose contents are bundled as package resources. Whereas `ModelsTests` and `TokenizersTests` each contain a `Resources` directory. This PR renames `FallbackConfigs` to `Resources` for consistency. It also moves the `Vocabs/bert-vocab.txt` file in `TokenizerTests` to that module's `Resources` directory.


FL33TW00D:
Solid clean up!

pcuenca:
I named it differently to better express intention and because these would ideally be removed at some point in the future. We shouldn't need them, as opposed to resources used elsewhere. It's fine to rename, no need to revert. In fact, I'll check if the reasons still stand or if we can remove.

Issue #176, opened 2025-02-12
Interested in building on Windows?


stevenbrix:
I'm curious how interested ya'll are on in accepting contributions to get this library building on Windows?

I've got it working in my fork here: https://github.com/huggingface/swift-transformers/compare/main...stevenbrix:swift-transformers:windows

Aside from compiling out some `CoreML` and `Accelerate` usage, it wasn't all that bad. I'm not sure what your feelings are on the changes made to `TensorUtils` which are making up for missing `Accelerate` xcframework APIs. I'm curious to get ya'lls thoughts. Thanks!

FL33TW00D:
Thanks for this @stevenbrix.

We've taken a look and I think this makes a strong case for splitting out more `Products`, since `CoreML` et al will never work on Windows (instead of flagging it out).

mattt:
Hi @stevenbrix, thanks for sharing that! We updated the sources to conditionally import platform-specific frameworks like CoreML and Accelerate (and FoundationNetworking, for that matter). Versions [1.1.7](https://github.com/huggingface/swift-transformers/releases/tag/1.1.7) and later support Linux. While we don't currently have plans to support Windows officially, I suspect that should mostly work right now.

PR #247, opened 2025-09-19
Conditionally import CoreML and Accelerate frameworks

mattt:
Related to https://github.com/huggingface/swift-transformers/issues/176

The `Hub` module and other products of the `swift-transformers` package could theoretically be used on Linux, Window, Wasm, or other platforms. As a prerequisite, this PR conditionalizes imports of `CoreML` and `Accelerate`, which are specific to macOS and iOS.

### DIFF
--- a/Sources/Generation/Generation.swift
+++ b/Sources/Generation/Generation.swift
@@ -5,7 +5,9 @@
 //  Created by Pedro Cuenca on 7/5/23.
 //
 
+#if canImport(CoreML)
 import CoreML
+
 import Tokenizers
 
 public enum GenerationMode {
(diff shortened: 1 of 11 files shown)


FL33TW00D:
LGTM!

pcuenca:
Can we potentially restrict `Generation` to Apple platforms?

mattt:
> Can we potentially restrict `Generation` to Apple platforms?

I think that's possible, but I don't think I've ever seen that in the wild. I'm familiar with how errors are communicated when APIs aren't available due to conditional compilation. If we wanted to make that experience nicer, we could add an `#else` condition that either 1) stubs non-extended declarations, like `GenerationMode`, with `@deprecated` notice or 2) emits a `#warning".

pcuenca:
Ok, no need to spend more time on it then. Thank you!

PR #216, opened 2025-08-21
Add VSCode settings and recommended extensions

mattt:
In working on #168, I found that my editor's configured format-on-save behavior was causing churn and [CI failures](https://github.com/huggingface/swift-transformers/actions/runs/17126880235/job/48580616211?pr=168#step:4:528) (sorry about that!)

This PR adds a `.vscode` directory to configure VSCode (plus derivatives like Cursor) to do the following:

- Recommend the installation of the [official Swift extension (`swiftlang.swift-vscode`)](https://marketplace.visualstudio.com/items?itemName=swiftlang.swift-vscode) and [`vknabel.vscode-swiftformat`](https://marketplace.visualstudio.com/items?itemName=vknabel.vscode-swiftformat)
- Configures format-on-save using SwiftFormat using the project's `.swiftformat` file

These settings should reduce the number of failed CI jobs due to formatting issues, and provide a nice baseline experience for developers using VSCode, et al.

PR #248, opened 2025-09-19
Migrate from SwiftFormat to swift-format

mattt:
SwiftFormat is a fantastic tool. But because it's not built-in to the Swift toolchain, adopting it for the project introduces friction. I've seen formatting issues creep up in several PR review threads (and not just from external contributors). To put some numbers behind that: GitHub says that the `format` workflow [fails 23% of the time](https://github.com/huggingface/swift-transformers/actions/metrics/performance) (compared to 35% for unit tests)

Me, I'll never remember to run this myself before committing changes, which is why I opened #216 to configure my editor to do this for me automatically. However, the suggested extension for SwiftFormat requires the user to have `swiftformat` available in `PATH`.

Even when `swiftformat` is available, it's difficult to manage versions — as we saw in #212. I attempted to solve that by managing and running SwiftFormat via Swift Package Manager, but I don't like the ergonomics (`swift package plugin --allow-writing-to-package-directory swiftformat --config .swiftformat .` 😵‍💫) and I _especially_ don't like how it introduces a transitive dependency to all downstream consumers.

So faced with these challenges, I propose that we go with the flow and standardize on `swift format`, which is built into the Swift toolchain.

I did my best to make `.swift-format` match the existing project settings as closely as possible. The resulting diff (97ad29f) is admittedly large, but I don't think it's too onerous (especially since we can ignore with `.git-blame-ignore-revs`). To my eyes, the formatting diff seems limited to:

1. Removing spaces in empty braces (`{ }` -> `{}`)
2. Indentation of chained methods and conditional statements

On the plus side, `swift format` has a `lint` subcommand that greatly simplifies our existing format checking workflow, and provides some helpful checks. For example, it flagged a few instances where documentation comments for a method were out of sync with the signature (e4f76dd8d20df9a07840b82ddf176c856d397161).

I understand and acknowledge that issues of project standards can be simultaneously contentious and banal. So I appreciate your taking the time to hear me out 🫶

### DIFF
--- a/Sources/Hub/HubApi.swift
+++ b/Sources/Hub/HubApi.swift
@@ -44,8 +44,8 @@ extension HTTPURLResponse {
 
             if trimmed.contains("rel=\"\(rel)\"") || trimmed.contains("rel=\(rel)") {
                 if let start = trimmed.firstIndex(of: "<"),
-                   let end = trimmed.firstIndex(of: ">"),
-                   start < end
+                    let end = trimmed.firstIndex(of: ">"),
+                    start < end
                 {
                     let startIndex = trimmed.index(after: start)
                     return String(trimmed[startIndex..<end])
@@ -368,7 +368,7 @@ public extension HubApi {
             let nextChunk = try? fileHandle.read(upToCount: chunkSize)
 
             guard let nextChunk,
-                  !nextChunk.isEmpty
+                !nextChunk.isEmpty
             else {
                 return false
             }
@@ -376,7 +376,7 @@ public extension HubApi {
             hasher.update(data: nextChunk)
 
             return true
-        }) { }
+        }) {}
 
         let digest = hasher.finalize()
         return digest.map { String(format: "%02x", $0) }.joined()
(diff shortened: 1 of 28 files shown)


FL33TW00D:
Tagging @ZachNagengast as he gave us good guidance on formatting last time.

100% agree these things are usually banal yet contentious. Any solution that works and causes minimal friction is fine with me.

Sources/Hub/HubApi.swift
```diff
@@ -298,9 +298,7 @@ public extension HubApi {
     ///
     /// Reference: https://github.com/huggingface/huggingface_hub/blob/b2c9a148d465b43ab90fab6e4ebcbbf5a9df27d4/src/huggingface_hub/_local_folder.py#L263
     ///
-    /// - Parameters:
-    ///   - localDir: The local directory where metadata files are downloaded.
-    ///   - filePath: The path of the file for which metadata is being read.
+    /// - Parameter metadataPath: The path of the metadata file to read.
```
pcuenca, line 301: lol nice

PR #249, opened 2025-09-19
Distinguish between configuration objects and values

mattt:
My understanding is that tokenizer configuration files distinguish between canonically equivalent strings (for example, `"a\u{300}"` and `"à"`). And so far as I can tell from the existing test coverage and discussion of #189, this is an important consideration.

The current implementation of `Config` makes a category error by conflating configuration objects with values. And in doing so, I believe it adds a lot of unnecessary work by introducing a `BinaryDistinctString` type to override the default `Swift.String` notion of equality (`==` implies canonical equivalence); this is notably different than `NSString`, which I believe compares UTF-16 code units).

Instead of doing all of that, this PR defines a narrow `Config.Key` type that encapsulates the desired semantics, and creating a distinct `Config.Value` enumeration that stores conventional `String` and `[String: Value]` values.

This looks more correct to me, but I don't have the full context for all of the decisions that led to the current design. So apologies if I'm making any obvious mistakes in my reasoning 🙇


mattt:
(For the record, I dislike the current `dynamicMemberLookup` behavior as I find it difficult to reason about, and would prefer that we deprecate that access pattern ahead of our 1.0 release)

pcuenca:
Yes, indeed, `NSString` and `String` are quite different: #113, #116. I'll review properly today.

Sources/Hub/Config.swift
```diff
@@ -3,811 +3,403 @@
...
-            return nil
         }
 
-        public func integer() -> Int? {
-            if case let .integer(val) = self {
-                return val
-            }
-            return nil
+        public static func == (lhs: Self, rhs: Self) -> Bool {
+            (lhs.value as NSString).isEqual(to: rhs.value)
```
pcuenca, line 31: Shouldn't the rhs be NSString as well?

Sources/Hub/HubApi.swift
```diff
@@ -269,7 +269,7 @@ public extension HubApi {
     func configuration(fileURL: URL) throws -> Config {
         let data = try Data(contentsOf: fileURL)
         let parsed = try JSONSerialization.bomPreservingJsonObject(with: data)
-        guard let dictionary = parsed as? [NSString: Any] else { throw Hub.HubClientError.parse }
+        guard let dictionary = parsed as? [String: Any] else { throw Hub.HubClientError.parse }
```
pcuenca, line 272: I'm afraid this is not going to work for vocab entries in tokenizers

pcuenca:
> (For the record, I dislike the current `dynamicMemberLookup` behavior as I find it difficult to reason about, and would prefer that we deprecate that access pattern ahead of our 1.0 release)

I'm not opposed to this. But I think that code that looks like `modelConfig.taskSpecificParams?.textGeneration?.textGenerationParameters?.doSample` looks more swift-y and less intimidating than the equivalent `modelConfig["task_specific_params"]...` chain. And if we were to parse to encodable types at some point, the result would resemble the former. One option is to just parse the entries we know about and ignore everything else, but I didn't know what we were going to need and left it open.

mattt:
Thanks for your review, @pcuenca. Closing this for now, since this is out of scope for v1. Hopefully we'll have a chance to revisit this later.

PR #250, opened 2025-09-19
Reduce access level for extensions on imported types

mattt:
This PR seeks to minimize the public API surface area of this package and avoid namespace pollution by reducing the access level for extensions on public types (like `String` and `Array`), insofar as they're being used as helpers.


Sources/Tokenizers/String+PreTokenization.swift
pcuenca, line 1: These are quite motivated by pre-tokenizer needs that mimic the Rust and JS implementations; not so much utility extensions per se.

Sources/Hub/HubApi.swift
```diff
@@ -892,13 +892,13 @@ public extension Hub {
     }
 }
 
-public extension [String] {
+private extension [String] {
```
pcuenca, line 894: why all the way to `private` in this case? (just curious)

Sources/Tokenizers/String+PreTokenization.swift
mattt: Gotcha. Yeah, Swift unfortunately lacks good options to selectively import API extensions to external types. Usually it's not a problem, but I've been bit before by a dependency polluting the standard library with APIs that conflicted with local extensions, which was confusing to sort out. I'm probably over-indexing on that experience, but my general rule is to avoid making these extensions public unless that's the intended purpose of the package.

Sources/Hub/HubApi.swift
mattt: Principle of Least Privilege. Minimizing the access level of APIs makes dependencies more explicit so you don't have dead code lying around. For example, if we refactored the methods in this file that use this extension on `[String]`, we know that we can safely remove the extension as well, because it can't be used anywhere else.

Sources/Tokenizers/String+PreTokenization.swift
pcuenca: Yes 👍 I was referring to the file name a bit, it doesn't convey a lot of meaning to name it like we name extension bags.

mattt: Ah, I see. Yeah, this could benefit from a more descriptive name. How about `String+Split.swift` or something like that? Or maybe `String+PreTokenization.swift`?

pcuenca: Yeah, `String+PreTokenization.swift` sounds good to me!

PR #253, opened 2025-09-23
Ignore formatting changes in `git blame`

mattt:
Follow up to #248
Related to #194 

This PR adds d907980, the squashed commit for #248, to `.git-blame-ignore-revs`, so that the formatting differences caused by running `swift format` don't cause an additional entry in a file's history when using `git blame` or GitHub's UI.


FL33TW00D:
Great stuff!

PR #257, opened 2025-09-24
Stateful cache, MLTensor

pcuenca:
This is a rebase of [the preview branch](https://github.com/huggingface/swift-transformers/tree/preview) on the current `main`.

What we gain:
- Modern `MLTensor` API
- Stateful models for easier KV caching
- Lots of tensor operation lines removed (no more `MLShapedArray`, `MLMultiArray` 🥳)
- A Mistral 7B conversion example

What we lose:
- The LogitsWarper abstraction that was adapted from Python. In particular:
  - Top-P sampling (#23, #27).
  - Repetition Penalty (#85).
- Transformers CLI to quickly test the converted models.

Some cleanup still needed.

Should we move forward with this? @mattt @FL33TW00D @Vaibhavs10 @ZachNagengast


Package.swift
```diff
@@ -10,7 +10,7 @@ let swiftSettings: [SwiftSetting] = [
 
 let package = Package(
     name: "swift-transformers",
-    platforms: [.iOS(.v16), .macOS(.v13)],
+    platforms: [.iOS("18.0"), .macOS("15.0")],
```
pcuenca: `.v18` and `.v15` not supported for manifests < Swift 6

pcuenca: We could revert and conditionally exclude the Generation target instead

FL33TW00D:
This is great! Cuts out a lot of cruft we don't need any more and we can repost about the Mistral 7B example (I still use it as a reference and even I struggle to find it!!)

pcuenca:
If we go forward, we can:

- [ ] Try to quickly add support for top-p sampling quickly. This way, what we'd be missing is:
  - Repetition penalty
  - Combined warpers (you can't use two of them simultaneously)
- [x] Move the CLI to the examples folder ~~https://github.com/huggingface/swift-chat or to https://github.com/huggingface/coreml-examples. This is not critical for release~~.

Examples/transformers-cli/Package.swift
```diff
@@ -0,0 +1,22 @@
+// swift-tools-version: 6.2
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "transformers-cli",
+    platforms: [.iOS(.v18), .macOS(.v15)],
+    dependencies: [
+        .package(url: "https://github.com/huggingface/swift-transformers", branch: "main"),
```
pcuenca:
```suggestion
        .package(url: "https://github.com/huggingface/swift-transformers", from: "1.0.0"),
```

to update

mattt: We can use a relative import here

```suggestion
        .package(path: "../../"),
```

Examples/transformers-cli/Sources/transformers-cli/Transformers.swift
```diff
@@ -0,0 +1,141 @@
+import ArgumentParser
+import CoreML
+import Foundation
+import Generation
+import Models
+
+@available(macOS 15.0, iOS 18.0, *)
```
mattt, line 7: Do we need this annotation if the package already defines these supported platform?

pcuenca: We don't, forgot to remove them. Thanks for pointing out!

pcuenca: Actually, I ended up not imposing the restriction on the package itself (i.e., preserve what's currently in main), but restrict Core ML features with availability annotations.

PR #259, opened 2025-09-24
Seed `NetworkMonitor` state from `currentPath` to eliminate offline race at startup

mattt:
Follow-up to #229 

@pcuenca observed that tests were failing locally because `shouldUseOfflineMode` was occasionally evaluating `isConnected` to be `false`. This new behavior occurred after switching to Swift Testing, in #229. That PR didn't change anything about offline mode, but it did increase test parallelism, which I suspect surfaced a race condition in the network monitor.

This PR updates the network monitor to seed the offline / online state immediately from `NWPathMonitor.currentPath` after calling `startMonitoring`. This should fix intermittent offline-mode activation that's observed running tests locally.


mattt:
Huh, CI is failing with a new test failure:

> /Users/runner/work/swift-transformers/swift-transformers/Sources/Hub/HubApi.swift:503: error: -[HubTests.SnapshotDownloadTests testCustomEndpointDownload] : failed: caught error: "unexpectedError"

Is this a flake? Or did we uncover a different race condition?

Sources/Hub/HubApi.swift
```diff
@@ -801,6 +801,13 @@ extension HubApi {
             }
```
ZachNagengast, line 801: Noticed that `startMonitoring` gets called twice using the singleton method `NetworkMonitor.shared.startMonitoring()` if one doesnt yet exist. May want to also add a check for `monitor.pathUpdateHandler == nil` before setting it here in case the first update goes to a handler that was already replaced.

Issue #251, opened 2025-09-22
Cleanup: throw error on missing `tokenizer_config.json`?


pcuenca:
Context: https://github.com/huggingface/swift-transformers/pull/246#issuecomment-3311748056

Some repos still don't have a `tokenizer_config.json` file ([example](https://huggingface.co/google-t5/t5-base/tree/main)), and it's not going to be added for BC considerations.

Question: should we just throw when this happens?

More info:

- Loading the tokenizer with transformers works. Saving it generates all necessary files:
```py
>>> from transformers import AutoTokenizer
>>> t = AutoTokenizer.from_pretrained("t5-base")
>>> t.save_pretrained("t5_saved")
```

- transformers.js raises an exception. But since it relies on onnx repos that have been previously exported, it's a non-issue – exporting goes through the previous path and the file is available, see [Xenova/t5-base](https://huggingface.co/Xenova/t5-base/tree/main)

Thoughts on doing the same (throw an error) for the sake of cleanup?

cc @mattt @FL33TW00D

mattt:
@pcuenca Repeating back my own understanding to clarify: 

`AutoTokenizer.from(pretrained:)` currently throws if configuration isn't found.

https://github.com/huggingface/swift-transformers/blob/a814cc530be5d1f847b532b1a53e54147a149e9a/Sources/Tokenizers/Tokenizer.swift#L629-L639

...but the fallback logic in `LanguageModelConfigurationFromHub` prevents that from happening for T5, specifically:

https://github.com/huggingface/swift-transformers/blob/a814cc530be5d1f847b532b1a53e54147a149e9a/Sources/Hub/Hub.swift#L105-L128

https://github.com/huggingface/swift-transformers/blob/a814cc530be5d1f847b532b1a53e54147a149e9a/Sources/Hub/Hub.swift#L242-L262

Therefore, you're proposing to remove that fallback and the bundled resources so that `AutoTokenizer.from(pretrained:)` starts throwing an error for T5, correct?

I don't have enough context to understand the impact of this, but I'm generally in favor of matching the behavior of sibling projects and reducing the size of our resources.

pcuenca:
Yes, that's correct.

Sibling projects work differently - transformers works (it has code to deal with these cases), transformers.js doesn't. In general, it's a good goal to be able to instantiate a tokenizer from any transformers model on the Hub, but these are old formats easy to convert.

mattt:
[...]

PR #261, opened 2025-09-24
Tokenizers don't need config.json

pcuenca:


pcuenca:
I'll add a test later.

pcuenca:
I tried to skip downloads of model configurations, but I realized there are still edge cases for old tokenizers where the [tokenizer class is retrieved from the model config](https://github.com/huggingface/swift-transformers/blob/3f1a93f363799726cac925423b9fddb614003118/Sources/Hub/Hub.swift#L120). We need to look into this better. This is another fallback mechanism in addition to #251.

pcuenca:
Merging this for now, will improve the fallback cases later.

PR #240, opened 2025-09-16
Upgrade Jinja package to version 2.0.0

mattt:
Related to https://github.com/huggingface/swift-jinja/pull/26

PR #262, opened 2025-09-25
Lint

pcuenca:


mattt (approved):

mattt:
@pcuenca I'm fighting CI in #240, too. I'm not sure what changed from under our feet to cause lint to start failing. It's frustrating that the `pull_request_target` is causing the workflow in `main` to be run instead of the updated one here. Not sure what the right fix is there. (YOLO?)

pcuenca:
Yes, I'm seeing these problems since yesterday night. In addition, `swift format` was apparently randomly failing to update my files locally for some reason (might be user error).

I think we only need `pull_request_target` for the external workflow, but for now we can merge this or #240 and verify.

mattt:
Extremely spooky that nothing seems to have changed in CI runner or our workflow, but we started getting failures...

Last succeeded: https://github.com/huggingface/swift-transformers/actions/runs/17984247022/job/51158207015
First failed: https://github.com/huggingface/swift-transformers/actions/runs/18005981937/job/51226463719?pr=262


Both have identical output for initial setup job step:

Screenshot 2025-09-25 at 04 36 02

mattt:
> I think we only need `pull_request_target` for the external workflow, but for now we can merge this or #240 and verify.

Let's go ahead and merge this, since it's a more targeted change, and then I'll try again on #240

mattt:
@pcuenca 🙃 ✅ https://github.com/huggingface/swift-transformers/actions/runs/18006286755/job/51227165617?pr=240

pcuenca:
I'll wait for #240 and then I'll do #261 and #257.

PR #264, opened 2025-09-25
Fix Swift 6 concurrency errors

mattt:
This PR fixes Swift 6 compiler warnings by adding `Sendable` conformance and `@MainActor` annotations to resolve new compiler errors that arise when building the package in Swift 6 language mode. Along the way, it fixes an existing data race in `PreTrainedTokenizer`.


mattt:
Before we merge, we should consider cutting a new minor release to lock in all of the changes we have so far, before 1.0.0.

PR #265, opened 2025-09-25
Change `TokenLattice` access level from public to internal

mattt:
Related to #243 and #250

`TokenLattice`, like `Trie`, is an implementation detail of `UnigramTokenizer`, and probably shouldn't be part of the public API. This PR changes its access level to `internal` accordingly.


pcuenca (approved):
Indeed, thanks!

PR #270, opened 2025-09-25
feat: tokenizers and hub are the big sellers!

FL33TW00D:

PR #272, opened 2025-09-26
Fix doc for `callAsFunction` that got out of sync

pcuenca:
I saw this during the lint run for #270, but tests passed when I merged #257 🤷


pcuenca:
I fixed the warnings (except `PresentKeys`, we are not dropping `Keys` here), but the CI does not seem to see my changes.

PR #273, opened 2025-09-28
Logit warpers

kashif:
fixes issue #271

PR #274, opened 2025-09-28
Re-implement Top-P sampling and `LogitsWarper` using `MLTensor`

### LINKED ISSUE #271: Add back Top-P sampling, Repetition Penalty, Logits Warpers
These are available in pre-1.0 releases, but were temporarily removed during the adaptation to `MLTensor` #257. Pull requests #23, #27, #85 can be used for reference.

mattt:
Resolves #271

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -33,7 +33,7 @@
         "AvoidRetroactiveConformances": true,
         "BeginDocumentationCommentWithOneLineSummary": false,
         "DoNotUseSemicolons": false,
-        "DontRepeatTypeInStaticProperties": true,
+        "DontRepeatTypeInStaticProperties": false,
         "FileScopedDeclarationPrivacy": true,
         "FullyIndirectEnum": true,
         "GroupNumericLiterals": false,
--- a/Sources/Models/LanguageModel.swift
+++ b/Sources/Models/LanguageModel.swift
@@ -129,7 +129,7 @@ extension LanguageModel {
         static let valueCache = "valueCache"
         // Output keys
         static let logits = "logits"
-        static let presentKeys = "presentKeys"
+        static let present = "presentKeys"
         static let presentValues = "presentValues"
     }
 }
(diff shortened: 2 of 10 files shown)


.swift-format
```diff
@@ -33,7 +33,7 @@
         "AvoidRetroactiveConformances": true,
         "BeginDocumentationCommentWithOneLineSummary": false,
         "DoNotUseSemicolons": false,
-        "DontRepeatTypeInStaticProperties": true,
+        "DontRepeatTypeInStaticProperties": false,
```
mattt, line 36: Strange that the [lint check fails](https://github.com/huggingface/swift-transformers/actions/runs/18075600846/job/51431586406?pr=274) even when this is disabled. FWIW, I also can't reproduce the lint failure locally...

> Sources/Models/LanguageModel.swift:132:20: warning: [DontRepeatTypeInStaticProperties] remove the suffix 'Keys' from the name of the variable 'presentKeys'

Sources/Models/LanguageModel.swift
```diff
@@ -129,6 +129,7 @@ extension LanguageModel {
         static let valueCache = "valueCache"
         // Output keys
         static let logits = "logits"
+        // swift-format-ignore: DontRepeatTypeInStaticProperties
```
mattt, line 132: Lint check failing despite disabling rule globally in `.swift-format` and using this `swift-format-ignore:` directive...

kashif:
@mattt  feel free to use any insights from my PR #273

mattt:
@kashif Ha! I just saw your PR opened at the same time and was about to comment the same 😃 

I haven't dug in too much, but on first glance, I think #273 has a better top-p and repetition penalty implementations that make better use of `MLTensor` functionality. I'll close mine for now so we can focus on yours instead. Thanks for your contribution!

kashif:
@mattt  no worries! I am test via:
```
swift run --package-path ../transformers-cli transformers-cli  --do-sample true --top-k 20 --top-p 0.9 --temperature 0.7 --repetition-penalty 1.5     \
  "Best recommendations for a place to visit in Paris in August 2024:" \
  --max-length 356 StatefulMistral7BInstructInt4.mlpackage
```

so you can also try that out

PR #276, opened 2025-09-29
Fix build on x86_64

Lakr233:
This pull request makes a targeted fix to the conditional compilation logic for attention mask handling in the `LanguageModelWithStatefulKVCache` class. The change ensures that the correct environment check is used for macOS and Mac Catalyst on x86_64 architecture.

* Compilation condition update: Replaced the incorrect use of `macCatalyst` with the proper `targetEnvironment(macCatalyst)` in the `#if` directive, ensuring accurate platform detection for attention mask logic in `LanguageModelWithStatefulKVCache`.

### DIFF
--- a/Sources/Models/LanguageModel.swift
+++ b/Sources/Models/LanguageModel.swift
@@ -471,7 +471,7 @@ public class LanguageModelWithStatefulKVCache: LanguageModel {
             Keys.inputIds: inputIds
         ]
         if isRequiringAttentionMask {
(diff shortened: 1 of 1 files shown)


Lakr233:
Fix #275

mattt:
@Lakr233 This is great! Thanks so much for reporting #275 and providing a fix.

Apologies for our busted CI. This change looks good to me :shipit:

mattt (approved):

PR #279, opened 2025-10-05
Updated @available for other platforms

jkrukowski:
`MLTensor` is available for tvOS 18.0+, visionOS 2.0+ and watchOS 11.0+ (see the docs [here](https://developer.apple.com/documentation/coreml/mltensor)).
This has to be specified explicitly in already existing `@available` attributes in order to make it compile on the mentioned platforms.

### DIFF
--- a/Sources/Generation/Decoders.swift
+++ b/Sources/Generation/Decoders.swift
@@ -3,7 +3,7 @@ import CoreML
 
 // MARK: Greedy Decoding
 
-@available(macOS 15.0, iOS 18.0, *)
+@available(macOS 15.0, iOS 18.0, tvOS 18.0, visionOS 2.0, watchOS 11.0, *)
 func selectNextTokenUsingGreedyDecoding(from scores: MLTensor) -> MLTensor {
     let indices = scores.argmax(alongAxis: -1).reshaped(to: [1, 1])
     // Ensure indices are Int32 for concatenation with input tokens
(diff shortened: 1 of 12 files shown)


mattt (commented):
Hi @jkrukowski. Thanks for opening this PR!

The package itself [targets only iOS and macOS](https://github.com/huggingface/swift-transformers/blob/4870d18b9f7637785891ab3e4c28f3ea489723d7/Package.swift#L13C5-L13C43), so I'm not sure expanding these `@available` annotations to include these unsupported platforms will have the desired effect.

```swift
let package = Package(
    name: "swift-transformers",
    platforms: [.iOS(.v16), .macOS(.v13)],
```

Curious to know: Are you developing for any of these platforms specifically and think we should update the package to target them as well?

jkrukowski:
hi @mattt thanks for the answer.
AFAIK specifying platforms in Package.swift does not restrict the package usage only to the specified platforms. It still allows the package to be used elsewhere but the minimum compatible version is provided implicitly, see the comment here https://forums.swift.org/t/swift-package-manager-and-supported-platforms-avoiding-available-macos-10-15-0/59882/2

I'm the creator of [swift-embeddings](https://github.com/jkrukowski/swift-embeddings) and after updating to `swift-transformers` 1.0 my package stopped building on `watchOS` and `tvOS` and expanding `@available` annotations (as done in the PR) fixes that. Let me know what you think

mattt:
@jkrukowski Ah, thanks for clarifying! In that case, I agree that this is the correct change to make.

PR #281, opened 2025-10-12
Fix offline mode check

DePasqualeOrg:
This fixes the logic when `useOfflineMode` is set to false on `HubApi`. Previously it would not allow downloads over a cell connection even when `useOfflineMode` was set to false.


pcuenca:
Superseded by #286 (added you as a co-author there)

Issue #280, opened 2025-10-08
macOS 26 user unable to download models with error "offline mode: Repository not available locally"


ronaldmannak:
I received a bug report from a macOS 26 user. They are unable to download any model. They consistently get an `offline mode: Repository not available locally` error. The user is not using a firewall.

I haven't updated to macOS 26 yet and I'm unable to reproduce the issue on macOS 15 Sequoia. I tried to compile the again on Xcode 26 but that didn't solve their issue.

Has anyone seen this issue?

This is the code I'm using to download models:
```swift
  let modelId = "mlx-community/SmolLM-135M-8bit"
  let hub = RepoStore.load().hubAPI(downloadBase: "https://huggingface.co")  
  try await hub.snapshot(from: modelId) { progress in
    ...
  }
```
<img width="312" height="155" alt="Image" src="https://github.com/user-attachments/assets/af2426eb-e437-46b2-b842-242fb5bfcb65" />

jkrukowski:
Yes, I experience that as well after updating to `macOS 26`, will try to dig deeper into it

pcuenca:
Interesting, I'm on macOS 26 and haven't seen that. Will try to reproduce with that snippet.

mattt:
@ronaldmannak Could you check how you're calling that API?

```swift
let hub = RepoStore.load().hubAPI(downloadBase: "https://huggingface.co")  
```

The `downloadBase` parameter is the base directory for where to download (defaults to `~/Documents/huggingface`); `"https://huggingface.co"` is the default value for the `endpoint` parameter when `HF_ENDPOINT` isn't defined. There's at least one code path that throws an error if files aren't found in that directory, so that could be the cause.

pcuenca:
Indeed, I didn't notice you seem to want to use `endpoint` rather than `downloadBae` (unless `RepoStore` uses different semantics).

This works for me:

```swift
        let modelId = "mlx-community/SmolLM-135M-8bit"
        let hub = HubApi()
        let downloadedTo = try await hub.snapshot(from: modelId) { progress in
            print("Total Progress: \(progress.fractionCompleted)")
        }
```

ronaldmannak:
Sorry, my bad. I manually edited the code to shorten it and make it more readable, but accidentally mixed up the URLs. Here's the unedited code from the project. 
The weird globalActor is there so users can save the models on an external drive. This works fine on macOS Sequoia. 

```swift
[...]

PR #286, opened 2025-11-03
Optimistic default connectivity, respect useOfflineMode if set

ZachNagengast:
There are a few PR's relating to this issue already #281 #260, but this PR simply reverts the [change](https://github.com/huggingface/swift-transformers/commit/b9209b98077a1f8acd6db735c3e97e4f6fb4c659#diff-d008547ca96c12875517345e1c3ddd2cdae19582c8f414b275e8b29f10bf801fR477) that ignores explicitly setting `useOfflineMode` in constrained or expensive network conditions to its previous logic (noted here https://github.com/huggingface/swift-transformers/issues/280#issuecomment-3396036390). Now if `useOfflineMode` is set it will always be respected, and only if nil will check the monitor for the proper fallback behavior.

In addition to that, it relaxes the default connectivity to assume true until updated by the monitor. This will match the other defaults that were optimistic already:
```swift
/// Assume best case connection until updated by the monitor
public var isConnected: Bool = true <- updated from false
public var isExpensive: Bool = false
public var isConstrained: Bool = false
```

There is still the case where we need a faster initial update, which I believe #259 covers well.

Regarding `isExpensive` and `isConstrained`, I think they are less relevant to this library and would be better off left to the developer since both cases would still be able to download a model, only `isConnected` will require a different code path to potentially succeed - developers can make the decision with their own monitor and UI that would warn the user, e.g. give them the option to continue if they're ok with either case. Future work here could be surfacing that to developers without throwing when there technically is connectivity.

### DIFF
--- a/Sources/Hub/HubApi.swift
+++ b/Sources/Hub/HubApi.swift
@@ -575,7 +575,7 @@ public extension HubApi {
             .appendingPathComponent(".cache")
             .appendingPathComponent("huggingface")
             .appendingPathComponent("download")
-        
+
         let shouldUseOfflineMode = await NetworkMonitor.shared.state.shouldUseOfflineMode()
 
         if useOfflineMode ?? shouldUseOfflineMode {


Sources/Hub/HubApi.swift
```diff
@@ -576,7 +576,9 @@ public extension HubApi {
             .appendingPathComponent("huggingface")
             .appendingPathComponent("download")
 
-        if await NetworkMonitor.shared.state.shouldUseOfflineMode() || useOfflineMode == true {
+        let shouldUseOfflineMode = await NetworkMonitor.shared.state.shouldUseOfflineMode()
+
+        if useOfflineMode ?? shouldUseOfflineMode {
```
pcuenca, line 581: Yes, I think you and @DePasqualeOrg in #281 are correct here!

Sources/Hub/HubApi.swift
```diff
@@ -822,7 +825,7 @@ extension HubApi {
             if ProcessInfo.processInfo.environment["CI_DISABLE_NETWORK_MONITOR"] == "1" {
                 return false
             }
-            return !isConnected || isExpensive || isConstrained
+            return !isConnected
```
pcuenca, line 828: I don't have a strong opinion here, I see how it would be the responsibility of the user app whether to allow to proceed. But they can also force behaviour with `useOfflineMode`, can't they?

ZachNagengast: Yes, this is just a bit more lenient default, like @DePasqualeOrg mentioned the default when on a cell connection with expensive bandwidth is to throw, which would definitely be one way to surface the issue to the user - but I'm hoping this is a bit less opinionated / more lenient to reduce any surprises without solid documentation on when and where "non-fatal" network based throws will occur, i.e. the download is only technically impossible when there is no network connection at all.

pcuenca:
Merged, and added @DePasqualeOrg as a co-author as he first reported a solution in #281.

PR #291, opened 2025-11-23
Reduce high CPU usage during downloads

DePasqualeOrg:
I noticed that CPU usage in my app was above 300% during 4 concurrent model downloads. I made some changes that appear to reduce the CPU usage by about half:

- Bytes are appended to the buffer in batches of 16 kB instead of individually.
- `ContiguousArray<UInt8>` is used for collecting batches instead of `Array<UInt8>`.
- The unnecessary creation of a `TaskGroup` on each update was removed.

The use of `URLSession.bytes(for:)` is inherently expensive due to the processing of bytes one at a time in an `AsyncSequence`, but I'm not sure if there's a better alternative that allows for resumable downloads.


pcuenca (commented):
This looks ok to me. Do you have any feedback @ardaatahan?

PR #292, opened 2025-11-23
Handle Jinja error messages

DePasqualeOrg:
I'm trying to surface error messages from Jinja so that consuming apps can show them in the UI. Currently they're thrown as a `JinjaError`, which means consuming apps would have to check for that error type in order to show the error message. Here I'm checking for `JinjaError` in `applyChatTemplate` and wrapping the error message in `TokenizerError.chatTemplate`. I'm not sure if this is the best solution, but it does expose the error message from Jinja in my app.

See https://github.com/huggingface/swift-jinja/pull/39

### DIFF
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -787,41 +787,48 @@ public class PreTrainedTokenizer: @unchecked Sendable, Tokenizer {
             throw TokenizerError.missingChatTemplate
         }
 
-        let template = try compiledTemplate(for: selectedChatTemplate)
-        var context: [String: Jinja.Value] = try [
-            "messages": .array(messages.map { try Value(any: $0) }),
-            "add_generation_prompt": .boolean(addGenerationPrompt),
-        ]
-        if let tools {
-            context["tools"] = try .array(tools.map { try Value(any: $0) })
-        }
-        if let additionalContext {
-            // Additional keys and values to be added to the context provided to the prompt templating engine.
-            // For example, the app could set "tools_in_user_message" to false for Llama 3.1 and 3.2 if a system message is provided.
-            // The default value is true in the Llama 3.1 and 3.2 chat templates, but these models will perform better if the tools are included in a system message.
-            for (key, value) in additionalContext {
-                context[key] = try Value(any: value)
-            }
-        }
+        let renderedTemplate: String
+        do {
+            let template = try compiledTemplate(for: selectedChatTemplate)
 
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -17,15 +17,15 @@ let package = Package(
         .library(name: "Transformers", targets: ["Tokenizers", "Generation", "Models"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/huggingface/swift-jinja.git", from: "2.0.0")
+        .package(url: "https://github.com/DePasqualeOrg/Jinja.git", branch: "expose-error-descriptions")
```
mattt: Thanks again for your help with https://github.com/huggingface/swift-jinja/pull/39

For simplicity and flexibility, I'd like to keep swift-jinja dependency loose, as `from: 2.0.0`. Most API consumers will automatically upgrade to the latest, or otherwise have the opportunity to specify more recent version explicitly.


```suggestion
        .package(url: "https://github.com/huggingface/swift-jinja.git", from: "2.0.0")
```

The existing code `(error as? LocalizedError)?.errorDescription ?? "\(error)"` will behave correctly for versions before and after that PR.

DePasqualeOrg: Yes, this was just for testing, of course. I've removed the commit with the changes in Package.swift.

mattt:
Thank you for your contributions to this package and swift-jinja. I agree with your changes here.

@pcuenca Does this look good to you? Or do you have any suggestions for how to surface these errors differently?

pcuenca:
We could potentially use a different error, but this works for me if it works for @DePasqualeOrg!

pcuenca (approved):

DePasqualeOrg:
@pcuenca, @mattt, this can be reverted, since I verified that the error is handled correctly here: https://github.com/DePasqualeOrg/swift-transformers/tree/revert-jinja-error-handling

I think that while I was testing the error propagation with changes in swift-transformers and swift-jinja, something went wrong and the error was not handled as expected, which led me to conclude that we need to handle `JinjaError` explicitly. In fact, we don't, since it's now a `LocalizedError`.

pcuenca:
@DePasqualeOrg not a fan of touching history unless necessary, but will merge #293 as you prefer it, (and the code is a bit easier). I cherry-picked the commit from your repo to correctly attribute the fix.

I'll also create a release with the latest commits.

DePasqualeOrg:
Thank you. It's not a matter of preference, but of correctness. The check is not necessary, so we should remove it.

PR #295, opened 2025-12-02
Fix Hub API URL construction

DePasqualeOrg:
It's safer to use `appending(path:)` or `appending(component:)` instead of string interpolation to construct URLs. The latter properly encodes the slash in revisions like `pr/1`, which would otherwise result in a 404 error.

Before: `pr/1` → two path segments → 404
After: `pr/1` → `pr%2F1` → works

I've also replaced the use of the deprecated `appendingPathComponent` in the same file with `appending(path:)`.

### DIFF
--- a/Sources/Hub/HubApi.swift
+++ b/Sources/Hub/HubApi.swift
@@ -152,8 +152,8 @@ private extension HubApi {
                     )
                 }
             },
-            { try? String(contentsOf: .homeDirectory.appendingPathComponent(".cache/huggingface/token"), encoding: .utf8) },
-            { try? String(contentsOf: .homeDirectory.appendingPathComponent(".huggingface/token"), encoding: .utf8) },
+            { try? String(contentsOf: .homeDirectory.appending(path: ".cache/huggingface/token"), encoding: .utf8) },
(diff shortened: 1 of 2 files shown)


Sources/Hub/HubApi.swift
```diff
@@ -258,7 +258,12 @@ public extension HubApi {
     /// - Throws: HubClientError if the repository cannot be accessed or parsed
     func getFilenames(from repo: Repo, revision: String = "main", matching globs: [String] = []) async throws -> [String] {
         // Read repo info and only parse "siblings"
-        let url = URL(string: "\(endpoint)/api/\(repo.type)/\(repo.id)/revision/\(revision)")!
+        let url = URL(string: endpoint)!
+            .appending(path: "api")
+            .appending(path: repo.type.rawValue)
```
mattt, line 263: Just checking, this is the same as `description`, yeah? (what's currently used in string interpolation)

DePasqualeOrg: Yes, that seems to be the case. Using `.rawValue` explicitly makes the `String` type clear and avoids relying on the default string interpolation behavior matching the case name.

mattt: Perfect. I see that we use that elsewhere, so I'm glad to have more consistency.

PR #303, opened 2025-12-27
Optimizations for significantly faster tokenizer loading

DePasqualeOrg:
Currently, tokenizer loading is a major performance bottleneck in Swift, typically taking ~1400 ms compared to ~100 ms in Python.

This PR optimizes tokenizer loading for a **3.6x speedup**, saving ~1020 ms. [#302](https://github.com/huggingface/swift-transformers/pull/302) and [#304](https://github.com/huggingface/swift-transformers/pull/304) in swift-transformers and [#21](https://github.com/huggingface/swift-huggingface/pull/21) in swift-huggingface add further performance gains.

## Performance

| Metric | Before | This PR | Speedup |
|--------|--------|-------|---------|
| Load time | 1411ms | 391ms | **3.6x (1020 ms)** |

Tested with Qwen/Qwen3-0.6B-Base tokenizer (150k vocab, 150k merges) on M3 MacBook Pro.

A benchmark is included in `LoadingBenchmarks.swift`, which can be removed before merging or kept to measure the impact of future changes. Run it with `swift test --filter LoadingBenchmarks`.

## Problems with the Current Implementation

1. **Config wrapper overhead**: `Config.convertToBinaryDistinctKeys()` recursively wraps every JSON value. For 150k vocab + 150k merges, this means 300k+ object allocations taking ~1 second — only to be immediately unwrapped.

2. **Expensive merge lookups**: `[BytePair: Int]` uses string hashing for merge rank lookups, which is slow for 150k entries.

3. **Sequential initialization**: Expensive dictionary building happens sequentially, leaving CPU cores idle.

### 1. Config Bypass (~1000ms saved)
Extract vocab/merges directly from raw JSON before `Config` conversion, passing them to new fast-path initializers.

### 2. Integer-Packed Merge Keys (~180ms saved)
Replace `[BytePair: Int]` (expensive string hashing) with `[UInt64: Int]` (fast integer hashing):
```swift
// Pack two token IDs into one UInt64
let key = UInt64(tokenIdA) << 32 | UInt64(tokenIdB)
```

### 3. Parallel Dictionary Building (~106ms saved)
Use `async let` to build dictionaries concurrently:
```swift
// Phase 1: Independent tasks
async let tokensToIdsTask = buildTokensToIds(...)
async let mergesTask = parseMerges(...)

// Phase 2: Dependent tasks (after Phase 1 completes)
async let bpeRanksTask = buildBpeRanks(...)
async let idsToTokensTask = buildIdsToTokens(...)
```

### 4. Conditional `stringToId` for Unicode Edge Cases
Added an optional `stringToId` fallback dictionary for tokenizers with Unicode edge cases (e.g., Gemma's BOM-prefixed tokens). Only built when needed — most tokenizers skip this entirely, saving ~50ms.

## Backward Compatibility

Standard usage benefits from the faster loading automatically:

```swift
let tokenizer = try await AutoTokenizer.from(pretrained: "model-name")
let tokenizer = try await AutoTokenizer.from(modelFolder: url)
```

Direct use of `LanguageModelConfigurationFromHub` continues to work unchanged. The default behavior preserves `tokenizerData.model.vocab` and `tokenizerData.model.merges` for backward compatibility.

For callers who want the performance optimization, opt in with `stripVocabForPerformance: true` and use the new properties:

```swift
let config = LanguageModelConfigurationFromHub(
    modelName: "model-name",
    stripVocabForPerformance: true
)
let vocab = try await config.tokenizerVocab   // NSDictionary for BPE
let merges = try await config.tokenizerMerges // [Any] for BPE
```

## Custom Tokenizer Registration

Added `AutoTokenizer.register(_:for:)` for registering custom tokenizer classes:

```swift
AutoTokenizer.register(MyTokenizer.self, for: "MyCustomTokenizer")
```

This mirrors Python transformers' [`AutoTokenizer.register()`](https://github.com/huggingface/transformers/blob/main/src/transformers/models/auto/tokenization_auto.py#L774-L804), which populates `REGISTERED_TOKENIZER_CLASSES` for lookup by class name.

This makes it easy for downstream projects like mlx-swift-lm to use the fast path via `AutoTokenizer.from()` while still supporting custom tokenizer classes when needed.

## Testing

All existing tests pass.

## Alignment with Python

These optimizations align with patterns in the Python tokenizers library:
- Packed merge keys ≈ tokenizers `Pair` type with tuple hashing
- Parallel building ≈ tokenizers Rayon-based parallelism
- Raw JSON extraction ≈ tokenizers `convert_to_native_format()`
- NSString for Unicode ≈ tokenizers byte-level preservation

## Future Work

A new major version could make breaking changes for better ergonomics:

- Async-first API: `Tokenizer.load(from:)` as primary entry point
- Factory methods instead of protocol-mandated initializers
- Hide Unicode complexity behind internal `TokenStorage` type

This would reduce BPETokenizer from 4 initializers to 1 factory method while maintaining performance.

### DIFF
--- a/Sources/Hub/Hub.swift
+++ b/Sources/Hub/Hub.swift
@@ -27,6 +27,18 @@ public enum TokenizerVocab: @unchecked Sendable {
     case unigram(NSArray)
 }
 
+/// Merge rules extracted from tokenizer.json for fast BPE initialization.
+///
+/// - Note: `@unchecked Sendable` is safe because the underlying data is immutable after extraction from JSON.
+public struct TokenizerMerges: @unchecked Sendable {
+    /// The raw merge rules as extracted from JSON.
+    public let rules: [Any]
+
+    public init(_ rules: [Any]) {
+        self.rules = rules
+    }
+}
+
 public extension Hub {
     /// Errors that can occur during Hub client operations.
     ///
@@ -142,7 +154,7 @@ public actor LanguageModelConfigurationFromHub {
     private var _tokenizerConfig: Config?
     private var _tokenizerData: Config?
     private var _tokenizerVocab: TokenizerVocab?
-    private var _tokenizerMerges: [Any]?
+    private var _tokenizerMerges: TokenizerMerges?
 
     /// Initializes configuration loading from a remote Hub repository.
     ///
@@ -257,8 +269,8 @@ public actor LanguageModelConfigurationFromHub {
         }
     }
 
-    /// Raw merges array extracted directly from JSON for fast BPE tokenizer initialization.
-    public var tokenizerMerges: [Any]? {
+    /// Merge rules extracted directly from JSON for fast BPE tokenizer initialization.
+    public var tokenizerMerges: TokenizerMerges? {
         get async throws {
             try await ensureLoaded()
             return _tokenizerMerges
@@ -281,7 +293,7 @@ public actor LanguageModelConfigurationFromHub {
         var tokenizerConfig: Config?
         var tokenizerData: Config
         var tokenizerVocab: TokenizerVocab?
-        var tokenizerMerges: [Any]?
+        var tokenizerMerges: TokenizerMerges?
     }
 
     /// Resolves tokenizerConfig with fallback logic.
@@ -373,7 +385,7 @@ public actor LanguageModelConfigurationFromHub {
 
             // Extract vocab/merges for fast tokenizer initialization (BPE and Unigram)
             var tokenizerVocab: TokenizerVocab? = nil
-            var tokenizerMerges: [Any]? = nil
+            var tokenizerMerges: TokenizerMerges? = nil
 
             if let modelDict = parsed["model"] as? NSDictionary {
                 let model = NSMutableDictionary(dictionary: modelDict)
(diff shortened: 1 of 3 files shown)


DePasqualeOrg:
https://github.com/huggingface/swift-transformers/pull/304 should be merged before this PR, which depends on it.

DePasqualeOrg:
@pcuenca, the performance numbers are specific to this PR, which was already stacked on https://github.com/huggingface/swift-transformers/pull/304 and has now been rebased on main after https://github.com/huggingface/swift-transformers/pull/304 was merged. My most recent tests actually show an even greater performance improvement than what I cited above.

I'm working on https://github.com/huggingface/swift-transformers/pull/311 as an alternative, which reduces the number of code paths at the expense of some breaking changes. I think some number of breaking changes are inevitable, so it makes sense to settle on an optimal design rather than supporting deprecated code paths, which make the code less readable.

I also think it makes sense to bundle these changes in one PR, since many are interlinked. The tokenizer registration changes would be easier to separate out, but https://github.com/ml-explore/mlx-swift-lm/pull/33 depends on this set of changes, and it would be nice to get this done in one go.

What are the missing edge cases that you mentioned?

DePasqualeOrg:

PR #308, opened 2026-01-29
Support compiling swift-transformers on Linux

Joannis:


pcuenca:
Interesting @Joannis! May I ask if you can share any specific uses you have for swift-transformers on Linux?

Joannis:
MLX-Swift

Sources/Hub/Hub.swift
```diff
@@ -49,29 +49,29 @@ public extension Hub {
         public var errorDescription: String? {
             switch self {
             case .authorizationRequired:
-                String(localized: "Authentication required. Please provide a valid Hugging Face token.")
```
Joannis, line 52: I hope these are not too controversial. If they are, I'll make them OS conditional

Sources/Hub/HubApi.swift
```diff
@@ -124,9 +131,22 @@ public struct HubApi: Sendable {
     /// The shared Hub API instance with default configuration.
     public static let shared = HubApi()
 
+    #if canImport(os)
     private static let logger = Logger()
+    #else
+    private static let logger = PrintLogger()
```
Joannis, line 138: Ideally we'd use `swift-log`. But I didn't want to add a dependency just yet. We can add it on Linux + Windows only if that's good.

Sources/Hub/HubApi.swift
```diff
@@ -124,9 +131,22 @@ public struct HubApi: Sendable {
...
+    #endif
 }
 
+#if !canImport(os)
+/// Simple print-based logger for non-Apple platforms
+private struct PrintLogger {
```
Joannis, line 144: Should be swift-log ideally

Sources/Hub/Hub.swift
pcuenca: This looks reasonable to me; however I have no idea if downstream projects use localized resources for these strings.

pcuenca: In any case, the end goal is to merge #297 when we can.

ronaldmannak: Yeah, they do. Or at least I do in my projects. :) 
I would appreciate if this could be conditional to keep localizations for Apple platforms

pcuenca: Thanks for the feedback @ronaldmannak 🙌

Sources/Hub/HubApi.swift
pcuenca: Yes, dependencies can be contentious so let's try to minimize if we can. Fine to add to Linux and Windows.

Joannis:
@pcuenca so to sum up; we need to revert localization on Darwin, and otherwise it's good right?

Examples/transformers-cli/Package.swift
```diff
@@ -11,6 +11,7 @@ let package = Package(
         // If you copy this manifest as a template, use the following line instead
         //.package(url: "https://github.com/huggingface/swift-transformers", from: "1.0.0"),
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.3.0"),
+        .package(url: "https://github.com/apple/swift-container-plugin", from: "1.0.0"),
```
pcuenca: Do we need this to run the example CLI?

Sources/Hub/HubApi.swift
```diff
@@ -69,7 +77,7 @@ extension HTTPURLResponse {
 /// and handling authentication with the Hugging Face Hub. It supports offline mode,
 /// background downloads, and automatic retry mechanisms for robust file transfers.
 public struct HubApi: Sendable {
-    var downloadBase: URL
+    public var downloadBase: URL
```
pcuenca: Is this public just so we can print it out from the example CLI?

Sources/Hub/HubApi.swift
```diff
@@ -300,13 +321,13 @@ public extension HubApi {
         public var errorDescription: String? {
             switch self {
             case let .invalidMetadataError(message):
-                String(localized: "Invalid metadata: \(message)")
+                ("Invalid metadata: \(message)")
```
pcuenca: Should we do the same trick here?

mattt:
@pcuenca Thanks, reviewing now

Examples/transformers-cli/Sources/transformers-cli-linux/main.swift
```diff
@@ -0,0 +1,227 @@
...
+import Foundation
+import Hub
+import Tokenizers
+
+/// Returns a HubApi configured to use persistent storage on WendyOS (/mnt/app),
+/// or the default location otherwise.
+func createHubApi() -> HubApi {
+    let wendyPersistentPath = "/mnt/app"
+    if FileManager.default.fileExists(atPath: wendyPersistentPath) {
+        let downloadBase = URL(filePath: wendyPersistentPath).appending(component: "huggingface")
```
mattt: I wasn't sure what WendyOS was until I found [this forum post](https://forums.swift.org/t/wendyos-a-swift-first-embedded-linux-distro/84478).  Since that's a not a requirement for Linux integration, I think we should keep this nonspecific.

```suggestion
/// Returns a HubApi configured to use persistent storage,
/// or the default location otherwise.
func createHubApi() -> HubApi {
    let root = "/mnt/app"
    if FileManager.default.fileExists(atPath: root) {
        let downloadBase = URL(filePath: root).appending(component: "huggingface")
```

mattt: Or better yet, since `/mnt/app` doesn't appear to be a conventional filesystem location, maybe we use the standard `~/.cache/huggingface/hub` base and add a flag to configure an alternative location (such `/mnt/app` for WendyOS)?

Examples/transformers-cli/Sources/transformers-cli-linux/main.swift
```diff
@@ -0,0 +1,227 @@
...
+        return HubApi(downloadBase: downloadBase)
+    }
+    return HubApi()
+}
+
+@main
+struct TransformersLinuxCLI: AsyncParsableCommand {
```
mattt: Is there any reason not to fold this into the existing CLI target?

Examples/transformers-cli/Package.swift
mattt: @pcuenca I don't believe so, no. Unless this gets referenced from a README for the example CLI as a "here's how you build a Swift container image", I think we should drop it.

```suggestion
```

Sources/Hub/Hub.swift
```diff
@@ -318,9 +318,22 @@ public final class LanguageModelConfigurationFromHub: Sendable {
...
+        // On non-Darwin platforms, try to locate resources relative to the executable
+        let executableURL = URL(filePath: CommandLine.arguments[0]).deletingLastPathComponent()
+        let possiblePaths = [
+            executableURL.appending(path: "swift-transformers_Hub.resources/\(modelType)_tokenizer_config.json"),
+            executableURL.appending(path: "../share/swift-transformers/\(modelType)_tokenizer_config.json"),
+        ]
+        guard let url = possiblePaths.first(where: { FileManager.default.fileExists(atPath: $0.path) }) else {
+            return nil
+        }
+        #endif
```
mattt, line 354: `modelType` comes from model config, so we need to treat it as untrusted data. Something like this will protect against path traversal attacks:

```suggestion
        let allowedModelTypeScalars = CharacterSet.alphanumerics.union(CharacterSet(charactersIn: "_-"))
        guard !modelType.isEmpty, modelType.unicodeScalars.allSatisfy(allowedModelTypeScalars.contains) else {
            return nil
        }
        let fallbackTokenizerConfigBaseName = "\(modelType)_tokenizer_config"
        let fileName = "\(fallbackTokenizerConfigBaseName).json"

        // Fallback tokenizer configuration files are located in the `Sources/Hub/Resources` directory
        // On Linux, Bundle.module may not be available if resources aren't properly bundled
        #if canImport(Darwin)
        guard let url = Bundle.module.url(forResource: fallbackTokenizerConfigBaseName, withExtension: "json") else {
            return nil
        }
        #else
        // On non-Darwin platforms, also try to locate resources relative to the executable
        var possiblePaths: [URL] = []
        if let executableDirectoryURL = Bundle.main.executableURL?.deletingLastPathComponent() {
            possiblePaths = [
                executableDirectoryURL
                    .appendingPathComponent("swift-transformers_Hub.resources", isDirectory: true)
                    .appendingPathComponent(fileName, isDirectory: false),
                executableDirectoryURL
                    .appendingPathComponent("..", isDirectory: true)
                    .appendingPathComponent("share", isDirectory: true)
                    .appendingPathComponent("swift-transformers", isDirectory: true)
                    .appendingPathComponent(fileName, isDirectory: false)
                    .standardizedFileURL,
            ]
        }
        guard let url = possiblePaths.first(where: { FileManager.default.fileExists(atPath: $0.path) }) else {
            return nil
        }
        #endif
```

Sources/Tokenizers/Utils.swift
mattt: All of these functions are unused, and should be removed outright.

I just opened https://github.com/huggingface/swift-transformers/pull/312 to do that.

Joannis:
I'll remove my test code and add CI

Joannis:
Not sure the lint check is correct. I feel it's ran over an old commit

.github/workflows/ci.yml
```diff
@@ -20,6 +20,31 @@ jobs:
...
+
+            - name: Build
+              run: swift build
+
+            - name: Run tests
+              env:
+                  CI_DISABLE_NETWORK_MONITOR: "1"
+                  HF_TOKEN: ${{ secrets.HF_HUB_READ_TOKEN }}
+              run: swift test
+
```
pcuenca, line 47: cc @glegendre01 for any potential security implications - Linux tests for this repo, in addition to the existing macOS ones we already had.

PR #311, opened 2026-02-11
Optimizations for significantly faster tokenizer loading (simplified)

DePasqualeOrg:
This is an experimental alternative to https://github.com/huggingface/swift-transformers/pull/303 that reduces code complexity at the expense of some breaking changes.


DePasqualeOrg:

PR #312, opened 2026-02-12
Remove unused utility methods and inline remaining contents of `Utils.swift`

mattt:
While reviewing #308, I noticed a change from `CFAbsoluteTimeGetCurrent` to `Date().timeIntervalSinceReferenceDate` that I wanted to double-check, and noticed that we aren't actually using that anywhere.

This PR removes this and the other unused utility methods in `Utils.swift`, and inlines the remaining ones.


pcuenca (approved):
Oh nice, thanks for taking the time to double-check and clean up!

PR #315, opened 2026-02-19
Replace `HubApi` downloader with `swift-huggingface` `HubClient`

mattt:
Alternative to #297

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -18,7 +18,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/huggingface/swift-jinja.git", from: "2.0.0"),
-        .package(url: "https://github.com/huggingface/swift-huggingface.git", from: "0.8.0"),
+        .package(url: "https://github.com/huggingface/swift-huggingface.git", from: "0.8.1"),
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.0.0"),
         .package(url: "https://github.com/apple/swift-crypto.git", "3.0.0"..<"5.0.0"),
         .package(url: "https://github.com/ibireme/yyjson.git", exact: "0.12.0"),


Sources/Hub/HubApi.swift
```diff
@@ -182,10 +243,175 @@ private extension HubApi {
...
+    }
+
+    func complete() {
+        lock.lock()
+        if progress.totalUnitCount <= 0 {
+            progress.totalUnitCount = 1
+        }
+        progress.completedUnitCount = progress.totalUnitCount
+        hasCompleted = true
+        lock.unlock()
```
pcuenca:
```suggestion
        hasCompleted = true
        lock.unlock()
        progress.completedUnitCount = progress.totalUnitCount
```

Setting `progress.completedUnitCount` acquires the lock via the `observation` progress handler. Not sure if this is the right thing to do, suggesting to move temporarily outside the lock so tests can proceed without deadlocks.

pcuenca:
My [download workaround](https://github.com/huggingface/swift-transformers/pull/315/commits/c0493bfc9469a9981864fdb4b2b5caddcf93c26d) fixes downloads for regular files but not for Xet ones.

I don't think we can solve this here. The problem, I think, is that [`copyFileToLocalDirectory` should receive a directory `destination`](https://github.com/huggingface/swift-huggingface/blob/7198b0633394b688859917249f1295542683790d/Sources/HuggingFace/Hub/HubClient%2BFiles.swift#L446) (not the full file path), while [`downloadFileWithXet` expects a file location](https://github.com/huggingface/swift-huggingface/blob/7198b0633394b688859917249f1295542683790d/Sources/HuggingFace/Hub/HubClient%2BFiles.swift#L458).

So we either have to update `copyFileToLocalDirectoryIfNeeded` to remove the trailing filename (and revert my commit here), or append the filename to `xetDestination`.

pcuenca:
The good news is that tests now proceed without deadlocks (after my second workaround) so we can see what's remaining.

Sources/Hub/HubApi.swift
```diff
@@ -182,10 +243,175 @@ private extension HubApi {
...
+                await Self.hubRepoIdCache.set(repo.id, value: canonical)
+                return canonical
+            }
+        }
+
+        // Keep historical behavior for non-qualified IDs when canonicalization fails.
+        return HuggingFace.Repo.ID(namespace: "", name: repo.id)
+    }
+
+    func makeHubClient(useBackgroundSession: Bool, enableCache: Bool = true) -> HubClient {
```
pcuenca: nit: maybe not call it `make`, as it returns a cached copy? (not a strong opinion though)

Sources/Hub/HubApi.swift
```diff
@@ -524,15 +756,12 @@ public extension HubApi {
...
-            if !FileManager.default.fileExists(atPath: incompleteDestination.path) {
-                try "".write(to: incompleteDestination, atomically: true, encoding: .utf8)
-            }
         }
 
         /// Downloads the file with progress tracking.
         /// - Parameter progressHandler: Called with download progress (0.0-1.0) and speed in bytes/sec, if available.
         /// - Returns: Local file URL (uses cached file if commit hash matches).
-        /// - Throws: ``EnvironmentError`` errors for file and metadata validation failures, ``Downloader.DownloadError`` errors during transfer, or ``CancellationError`` if the task is cancelled.
+        /// - Throws: ``EnvironmentError`` errors for file and metadata validation failures, ``Hub.HubClientError`` errors during transfer, or ``CancellationError`` if the task is cancelled.
```
pcuenca: should we consider this a small breaking change?

mattt: Good catch! Let's see if we can't keep `DownloadError` around and bridge to it from `HubClientError`.

mattt:
@pcuenca 

> My [download workaround](https://github.com/huggingface/swift-transformers/pull/315/commits/c0493bfc9469a9981864fdb4b2b5caddcf93c26d) fixes downloads for regular files but not for Xet ones.
> 
> I don't think we can solve this here. The problem, I think, is that [`copyFileToLocalDirectory` should receive a directory `destination`](https://github.com/huggingface/swift-huggingface/blob/7198b0633394b688859917249f1295542683790d/Sources/HuggingFace/Hub/HubClient%2BFiles.swift#L446) (not the full file path), while [`downloadFileWithXet` expects a file location](https://github.com/huggingface/swift-huggingface/blob/7198b0633394b688859917249f1295542683790d/Sources/HuggingFace/Hub/HubClient%2BFiles.swift#L458).
> 
> So we either have to update `copyFileToLocalDirectoryIfNeeded` to remove the trailing filename (and revert my commit here), or append the filename to `xetDestination`.

Whoops, sorry about that. I just opened a PR to fix upstream here: https://github.com/huggingface/swift-huggingface/pull/43

Sources/Hub/HubApi.swift
mattt: Fixed by [0a331ca](https://github.com/huggingface/swift-transformers/pull/315/commits/0a331ca9682d64f314a697ee7c250283fe7c6e57)

Sources/Hub/HubApi.swift
mattt: Fixed by 59e48d793d3f7edf685ef062d4f58fe206d86157

Package.swift
```diff
@@ -18,6 +18,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/huggingface/swift-jinja.git", from: "2.0.0"),
+        .package(url: "https://github.com/huggingface/swift-huggingface.git", from: "0.8.0"),
```
pcuenca:
```suggestion
        .package(url: "https://github.com/huggingface/swift-huggingface.git", from: "0.8.1"),
```

pcuenca (approved):
🔥

PR #324, opened 2026-03-02
Add JSON5 flag

ronaldmannak:
This PR enables JSON5 parsing of config.json. Some MLX models, like Nvidia Nemotron, use JSON5 formatting that fails to open  without the JSON5 flag being set.

### DIFF
--- a/Tests/TokenizersTests/YYJSONParserTests.swift
+++ b/Tests/TokenizersTests/YYJSONParserTests.swift
@@ -80,24 +80,13 @@ struct YYJSONParserTests {
     }
 
     @Test
-    func parsesJSON5Features() throws {
-        // Infinity and NaN (the primary motivation for enabling JSON5)
-        let jsonInfinity = #"{"limit": [0, Infinity], "neg": -Infinity, "nan_val": NaN}"#
-        let config = try YYJSONParser.parseToConfig(Data(jsonInfinity.utf8))
+    func parsesInfinityAndNaN() throws {
+        // YYJSON_READ_ALLOW_INF_AND_NAN enables Infinity, -Infinity, and NaN
+        let json = #"{"limit": [0, Infinity], "neg": -Infinity, "nan_val": NaN}"#
+        let config = try YYJSONParser.parseToConfig(Data(json.utf8))
         let limit = config["limit"].array()
         #expect(limit?[1].floating() == Float.infinity)
         #expect(config["neg"].floating() == -Float.infinity)
         #expect(config["nan_val"].floating()?.isNaN == true)
-
-        // Trailing commas in object and array
-        let jsonTrailing = #"{"a": 1, "b": [1, 2,],}"#
-        let config2 = try YYJSONParser.parseToConfig(Data(jsonTrailing.utf8))
-        #expect(config2["a"].integer() == 1)
-        #expect(config2["b"].array()?.count == 2)
-
-        // Single-line comments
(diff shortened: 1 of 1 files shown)


ronaldmannak:
Yes, this one for example: https://huggingface.co/mlx-community/NVIDIA-Nemotron-3-Nano-30B-A3B-4bit/blob/main/config.json

pcuenca:
Yes, I thought it would be related to the `Infinity` value that has given us problems before.

The long-term solution we are proposing for that case is to serialize [like this](https://github.com/huggingface/transformers/pull/42959/changes). Let me investigate internally a bit about how that is going.

Sources/Hub/YYJSONParser.swift
```diff
@@ -47,10 +47,11 @@ enum YYJSONParser {
             }
 
             var err = yyjson_read_err()
+            let flags: yyjson_read_flag = YYJSON_READ_JSON5
```
mattt: If we wanted to make the minimal change to support the `Infinity` in [mlx-community/NVIDIA-Nemotron-3-Nano-30B-A3B-4bit/config.json](https://huggingface.co/mlx-community/NVIDIA-Nemotron-3-Nano-30B-A3B-4bit/blob/main/config.json)

```suggestion
            let flags: yyjson_read_flag = YYJSON_READ_ALLOW_INF_AND_NAN
```

mattt:
@ronaldmannak Thanks for your PR!

I believe it's more accurate to say that those Nemotron models include the `Infinity` keyword, which is supported in Python but fails here because it's invalid in  strict JSON. JSON5 also supports the `Infinity` keyword, but I think its use here in `config.son` comes from a non-strict encoding from Python rather than an explicit decision by the model authors to use JSON5 (or at least I couldn't find any evidence to support that).

I think parsing as JSON5 is _probably_ fine. But I'd prefer a more targeted fix, like what I suggest in https://github.com/huggingface/swift-transformers/pull/324#discussion_r2871763100

ronaldmannak:
@mattt You are 100% right, `YYJSON_READ_ALLOW_INF_AND_NAN` is a better option. I didn't that existed. I've updated the code.
@pcuenca unit tests added but in a slightly different way. Lmk if this is okay with you.

Tests/TokenizersTests/YYJSONParserTests.swift
```diff
@@ -78,4 +78,26 @@ struct YYJSONParserTests {
             try YYJSONParser.parseToConfig(invalidJSON)
         }
     }
+
+    @Test
+    func parsesJSON5Features() throws {
+        // Infinity and NaN (the primary motivation for enabling JSON5)
```
pcuenca, line 84: Let's maybe remove the json5 references in the test name and comment since we are now targeting a more focused change :)

ronaldmannak: You might be looking at an older commit. The method was renamed to `parsesInfinityAndNaN` in https://github.com/huggingface/swift-transformers/pull/324/commits/7e847ede2405cc2db5d8d5142edb840f09322619

mattt:
Tests are passing locally and all PR feedback appears to be resolved, so I'm going to go ahead and merge this so that we can cut a patch release to unblock @davidkoski.

mattt:
This is now fixed in [1.1.9](https://github.com/huggingface/swift-transformers/releases/tag/1.1.9)

PR #305, opened 2026-01-11
Update Downloader.swift for background task support

musamuss:

PR #325, opened 2026-03-03
Harden legacy `Downloader` background behavior and progress reporting

mattt:
Most of the changes in #305 were superseded by our migration to [swift-huggingface](https://github.com/huggingface/swift-huggingface) in #315. However, there were some useful changes there that are worth pulling in, even if `Downloader` is only kept around for backward compatibility — specifically:

- Keeping background session identifiers collision-safe
- Guarding progress with unknown content-length
- Preserving temp files safely before move


mattt:
On second thought, `Downloader` is an internal type that's no longer used. I thought it still had some purchase in the public API, but I was mistaken.

Closing this.

PR #328, opened 2026-03-06
Fix yyjson parity with JSONSerialization

ZachNagengast:
We noticed a breaking change from #304  with one of our tokenizers that had trailing commas in it. 

`yyjson read failed (code 7) at position 118: trailing comma is not allowed.`

Luckily yyjson has a flag for this (expanding on @ronaldmannak's work in https://github.com/huggingface/swift-transformers/pull/324) so it can be brought back into parity with the previous JSONSerialization, but it requires a new release because protocolization was decided against.

This PR adds flags to reduce the validation error checks down to a level that JSONSerialization was allowing before.

Along with this, added some comprehensive tests to make sure we stay in parity with JSONSerialization in the future. 
Through writing these tests, I also found another breaking change regarding NUL characters where `String(cString: "before\u0000after")` would stop parsing after the NUL part, returning `before` only instead of `beforeafter`, but this was on the swift side not related to yyjson.

### DIFF
--- a/Sources/Hub/YYJSONParser.swift
+++ b/Sources/Hub/YYJSONParser.swift
@@ -52,12 +52,10 @@ enum YYJSONParser {
             //  - Inf/NaN literals in numeric fields
             //  - Trailing commas after the last element in objects/arrays
             //  - UTF-8 BOM prefix from files served over the network
-            //  - C-style single-line (//) and block (/* */) comments
             let flags: yyjson_read_flag =
                 YYJSON_READ_ALLOW_INF_AND_NAN
                 | YYJSON_READ_ALLOW_TRAILING_COMMAS
                 | YYJSON_READ_ALLOW_BOM
-                | YYJSON_READ_ALLOW_COMMENTS
             let doc = yyjson_read_opts(
                 UnsafeMutableRawPointer(mutating: baseAddress).assumingMemoryBound(to: CChar.self),
                 buffer.count,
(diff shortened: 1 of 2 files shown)


Sources/Hub/YYJSONParser.swift
```diff
@@ -47,7 +47,17 @@ enum YYJSONParser {
...
+            // tokenizer files that may contain non-standard JSON constructs:
+            //  - Inf/NaN literals in numeric fields
+            //  - Trailing commas after the last element in objects/arrays
+            //  - UTF-8 BOM prefix from files served over the network
+            //  - C-style single-line (//) and block (/* */) comments
+            let flags: yyjson_read_flag =
+                YYJSON_READ_ALLOW_INF_AND_NAN
+                | YYJSON_READ_ALLOW_TRAILING_COMMAS
+                | YYJSON_READ_ALLOW_BOM
+                | YYJSON_READ_ALLOW_COMMENTS
```
mattt: Do we have any examples in the wild of config files with comments? Maybe it's overly defensive, but I'd prefer to keep this narrowly tailored.

mattt: Removed this and the related tests with 63e0d322f8215f2ec1f67e9d5e7e5a1415810c72. Happy to revert/reapply that in a follow-up as needed.

ZachNagengast:
Good catch, appreciated the review!

PR #331, opened 2026-03-08
Add mixed-attention Core ML mask support for stateful generation

### LINKED ISSUE #330: Stateful Core ML generation does not support mixed-attention models that require multiple mask inputs
Summary

swift-transformers currently assumes a single-mask Core ML generation contract for stateful text generation.

This works for models that can be driven by one causal attention mask, but it breaks for stateful Core ML exports that require multiple mask inputs during generation, for example:
- one mask for full-attention layers
- one mask for sliding-window or local-attention layers

As a result, there is currently no stable way to use such stateful Core ML exports with the stock LanguageModel generation path.

Current runtime behavior

In Sources/Models/LanguageModel.swift, the stateful generation path only knows these input keys:
- inputIds
- attentionMask
- causalMask
- keyCache
- valueCache

Relevant locations:
- LanguageModel.Keys only defines attentionMask / causalMask
- predictNextTokenScores builds a single [1, 1, 1, tokenCount + 1] mask and passes that to the model

Why this is a problem

Some decoder-only Core ML exports use mixed attention:
- some layers use full attention
- some layers use sliding-window or local attention

Those exports may need multiple attention masks during generation instead of a single causal mask.

If the model exposes multiple attention-mask inputs, the current swift-transformers generation runtime cannot drive it.

Important scope clarification

This issue is specifically about supporting explicit multi-mask Core ML model contracts in the runtime.

A separate experimental exporter path that tried to reconstruct multiple masks inside the Core ML graph from a single causalMask turned out to be unstable in practice, with failures such as:

perm tensor length must equal input tensor rank, 4 != 8
'mps.tile' op input rank: 8 should match multiplier length: 4
original module failed verification

That exporter-side issue is separate. The request here is to support the stable explicit-input contract in the runtime.

Repro context

- Export target: stateful Core ML (iOS 18, StateType KV cache)
- Runtime: latest swift-transformers main
- swift-transformers commit tested: e5e227ba3225644d8a98c518eff8fb79be652223

Expected behavior

There should be a supported stateful Core ML generation path for decoder models that require multiple mask inputs.

Possible fix directions

1. Extend the runtime to support additional attention-mask inputs when they are present in modelDescription.

2. Add a more general mixed-attention Core ML generation path instead of assuming a single-mask contract.

3. If this is currently out of scope, document that the stock stateful Core ML generation API only supports single-mask causal-attention models.

Why this matters

Without this support, any mixed-attention Core ML export requires a custom runtime instead of the stock swift-transformers generation API.

Additional context

Example converted Core ML repo using the explicit multi-mask contract:
https://huggingface.co/Skyline23/translategemma-4b-it-coreml

Skyline-23:
What

Add support for stateful Core ML language models that require multiple attention masks during generation.

Why

The current runtime only handles attentionMask / causalMask, which is not sufficient for mixed-attention Core ML exports that need separate masks for different layer types.

This change allows the stateful generation path to populate:
- fullAttentionMask
- slidingAttentionMask

when those inputs are present in the Core ML model description.

Implementation

- add fullAttentionMask and slidingAttentionMask keys to LanguageModel.Keys
- detect those inputs from modelDescription
- build additive full-attention and sliding-window masks in the stateful generation path
- resolve the sliding window size from Core ML metadata first, and fall back to Hugging Face config when needed
- factor stateful generation input assembly into a reusable helper for test coverage
- keep existing single-mask models working unchanged

Tests

- add regression tests for additive full-attention mask construction
- add regression tests for sliding-window mask construction
- add an integration-style test that verifies the full input dictionary for a mixed-attention model contract
- verify with:
  swift test --filter LanguageModelCoreMLMaskTests

Scope clarification

This PR is intended to support explicit multi-mask Core ML generation contracts in the runtime.

It does not attempt to fix exporter-side approaches that reconstruct multiple masks inside a Core ML graph from a single causalMask input.

Additional context

Closes #330

Example converted Core ML repo using the explicit multi-mask contract:
https://huggingface.co/Skyline23/translategemma-4b-it-coreml

### DIFF
--- a/Sources/Models/LanguageModel.swift
+++ b/Sources/Models/LanguageModel.swift
@@ -139,6 +139,8 @@ extension LanguageModel {
         static let inputIds = "inputIds"
         static let attentionMask = "attentionMask"
         static let causalMask = "causalMask"
+        static let fullAttentionMask = "fullAttentionMask"
+        static let slidingAttentionMask = "slidingAttentionMask"
         static let keyCache = "keyCache"
         static let valueCache = "valueCache"
         // Output keys
@@ -260,6 +262,16 @@ public extension LanguageModel {
     /// Whether the model requires a causal attention mask.
     var isRequiringCausalMask: Bool {
         modelDescription.inputDescriptionsByName[Keys.causalMask] != nil
+    }
+
+    /// Whether the model requires a full-attention mask.
+    var isRequiringFullAttentionMask: Bool {
+        modelDescription.inputDescriptionsByName[Keys.fullAttentionMask] != nil
+    }
+
+    /// Whether the model requires a sliding-window attention mask.
+    var isRequiringSlidingAttentionMask: Bool {
+        modelDescription.inputDescriptionsByName[Keys.slidingAttentionMask] != nil
     }
 
     /// Determines the type of KV Cache available for the model, if any.
@@ -412,6 +424,27 @@ public extension LanguageModel {
         }
     }
 
+    /// Sliding window size for mixed-attention models, if present.
+    var slidingWindowSize: Int? {
+        get async throws {
+            if let userFields = metadata[MLModelMetadataKey.creatorDefinedKey] as? [String: String],
+                let value = userFields["co.huggingface.exporters.sliding_window"],
+                let parsed = Int(value)
+            {
+                return parsed
+            }
+
+            let modelConfig = try await modelConfig
+            if let direct = modelConfig?.slidingWindow.integer() {
+                return direct
+            }
+            if let nested = modelConfig?.textConfig.slidingWindow.integer() {
+                return nested
+            }
+            return nil
+        }
+    }
+
     /// The tokenizer associated with this model.
     ///
     /// Lazily loads and caches the tokenizer on first access.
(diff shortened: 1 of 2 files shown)


pcuenca (commented):
Very interesting and cool PR @Skyline-23! I won't be able to properly test and review it until the end of the week. Meanwhile, a couple of questions:

- The converted example model seems to be using float32 instead of float16 (because of [this line](https://huggingface.co/Skyline23/translategemma-4b-it-coreml/blob/main/convert_stateful_translategemma_coreml.py#L413), and because the repo takes ~16 GB). Did you try to convert to float16? Did you try any quantization options?
- Are you using or planning to use this Core ML model in a downstream app?

Thanks a lot for the contribution!

Skyline-23:
@pcuenca Sorry for late reply! It's fine. Please review slowly
The convert script was not optimized for the CoreML, I fixed it and add options with 4-bit and 8-bit quantized models.
It was problem with float 32 attention.
I will planning to use with on device translate app, In the test bed.
It could be canceled, but In my opinion this PR going to move repo good way for the various model support
Thanks!

john-rocky:
I gave this a local try on macOS 26 / M-series. PR builds cleanly against current `main` after a trivial rebase, the 4 new mask tests pass, and the full suite (120 tests) passes too.

I also pulled `Skyline23/translategemma-4b-it-coreml` (Int4 per-channel) to see the runtime path end-to-end. PR #331's detection works as expected on the loaded model:

```
isRequiringFullAttentionMask    = true
isRequiringSlidingAttentionMask = true
slidingWindowSize               = 1024
```

A greedy decode of 8 tokens after a 12-token prefill produces coherent output (≈ 2.3 tok/s, `.cpuAndGPU`). One thing I ran into: the `MLTensor`-based `model.prediction(from:using:)` overload that swift-transformers uses errored with *"The output feature named `logits`'s shape is data dependent and doesn't allow user specified output backing object."* on this model, so I ended up driving inference via `MLDictionaryFeatureProvider` directly. That's likely orthogonal to this PR, but flagging in case it's familiar.

Two small observations from testing:

**1.** The convert script doesn't write `co.huggingface.exporters.sliding_window` into the user-defined metadata. The fallback chain in `slidingWindowSize` (`modelConfig?.textConfig.slidingWindow`) recovers 1024 from Hub `config.json` for `google/translategemma-4b-it`, so it works on a developer machine with network. For fully offline loads or for users whose `co.huggingface.exporters.name` doesn't point to a public repo, the metadata key would be the more stable source. A one-line addition to `_apply_coreml_metadata` would close the gap:

```python
"co.huggingface.exporters.sliding_window": str(text_config.sliding_window),
```

**2.** Inside `statefulGenerationInputs`:

```swift
if includeSlidingAttentionMask {
    guard let slidingWindow else {
        fatalError(...)
    }
```

combined with `try? await slidingWindowSize` in `predictNextTokenScores`, a Hub-fetch failure (offline, rate-limited, gated repo) silently becomes `nil` and lands at this `fatalError`. Resolving the value once at init for stateful models that need it would surface the missing-config case earlier and more recoverably.

If a real-model integration fixture would be useful later, I'd be happy to convert and publish a small mixed-attention variant under my own namespace as a possible CI fixture — no need to block this PR on it.

PR #332, opened 2026-03-09
Add cache step to CI workflow

mattt:
Didn't notice until just now that we're not caching intermediate build output between CI runs. Especially since we brought in SwiftNIO / AsyncHTTPClient as a transitive dependency for Xet downloads, this adds an extra minute or two to each run.


pcuenca (approved):
These are in addition to the ones added in #334, right?

mattt:
@pcuenca Whoops, yeah — thanks for catching that. Looks like we should be covered now with #334.

PR #319, opened 2026-02-24
Fix MetaspacePreTokenizer: prepend_scheme no longer gated by addPrefixSpace

beshkenadze:
## Summary

- **Fixes** `MetaspacePreTokenizer.preTokenize()` which never prepended `▁` when `add_prefix_space` was absent from the tokenizer config, even when `prepend_scheme` was set to `"always"` — breaking XLM-RoBERTa and any SentencePiece Unigram model relying on Metaspace with `prepend_scheme: "always"`.
- **Aligns** with the canonical Rust implementation ([huggingface/tokenizers PR #1357](https://github.com/huggingface/tokenizers/pull/1357)) where `prepend_scheme` is the sole runtime authority for prepending behavior.
- **Adds** 9 new unit tests covering all `prepend_scheme` modes, backward compatibility with legacy `add_prefix_space`, precedence rules, and edge cases.

Fixes #318

## Root Cause

In the previous code, `addPrefixSpace` (defaulting to `false` when absent) was used as an outer gate for the `prependScheme` logic:

```swift
// BEFORE (broken):
if addPrefixSpace, !normalized.hasPrefix(replacement) {
    if prependScheme == .always { prepend = stringReplacement }
    if prependScheme == .first, options.contains(.firstSection) { prepend = stringReplacement }
}
```

When `add_prefix_space` was absent from the config (which is normal for XLM-RoBERTa), `addPrefixSpace` defaulted to `false`, and the entire prepend block was **never executed** — regardless of `prependScheme`.

### `Sources/Tokenizers/PreTokenizer.swift`

**`init`** — `prependScheme` now resolves with backward compat:
- If `prepend_scheme` is explicit in config → use it directly
- Otherwise derive from `add_prefix_space` (`true`/absent → `.always`, `false` → `.never`)

**`preTokenize`** — removed `addPrefixSpace` gate, replaced with `switch prependScheme`:
```swift
// AFTER (fixed):
if !normalized.hasPrefix(replacement) {
    switch prependScheme {
    case .always: prepend = stringReplacement
    case .first:
        if options.contains(.firstSection) { prepend = stringReplacement }
    case .never: break
    }
}
```

### `Tests/TokenizersTests/PreTokenizerTests.swift`

9 new test functions:

| Test | Validates |
|---|---|
| `always` without `add_prefix_space` | The XLM-RoBERTa fix (issue #318) |
| `first` with section options | First-section-only prepending |
| `never` | No prepending |
[...]

PR #345, opened 2026-05-01
Honor `prepend_scheme` in MetaspaceDecoder

### LINKED ISSUE #329: T5 Tokenizer prepending extra space on decode
When decoding a response from a fine-tuned T5 tokenizer, a leading space is prepended.

```
tokenIds: [571, 33, 25, 58]
tokenStrings ["▁How", "▁are", "▁you", "?"]
decoded [" How", " are", " you", "?"]
```

The issue is probably how the MetaspaceDecoder is initialized:

https://github.com/huggingface/swift-transformers/blob/e5e227ba3225644d8a98c518eff8fb79be652223/Sources/Tokenizers/Decoder.swift#L240

Or possibly how the MetaspaceTokenizer is initialized and handles legacy keys:

https://github.com/huggingface/swift-transformers/blob/e5e227ba3225644d8a98c518eff8fb79be652223/Sources/Tokenizers/PreTokenizer.swift#L185

Newer versions of python transformers are dropping/rewriting keys in the `tokenizer.json`. The decoder looks only for `addPrefixSpace` which is never derived from `prepend_scheme` when a new tokenizer config is specified.

I don't know the correct derivation, or the general plans for tokenizer migration, so hopefully this is enough information to make the correct fix. The python version of T5Tokenizer has some fallback code on initialization which obscures the data model change, but it wasn't clear if that appropriate here.

This will generate a tokenizer that shows the problem - this is transformers-5.3.
```
mpath = "google-t5/t5-small"
tokenizer = T5Tokenizer.from_pretrained(mpath)
tokenizer.save_pretrained("./t5-fine-tune")
```

Here's the key changes:
```
> diff -u <(jq -S . <./google-t5/t5-small/tokenizer.json) <(jq -S . < ./t5-fine-tune/tokenizer.json) 
--- /dev/fd/63	2026-03-07 06:55:12
+++ /dev/fd/62	2026-03-07 06:55:12
@@ -929,12 +929,14 @@
     }
   ],
   "decoder": {
-    "add_prefix_space": true,
+    "prepend_scheme": "always",
     "replacement": "▁",
-    "str_rep": "▁",
+    "split": true,
     "type": "Metaspace"
   },
   "model": {
+    "byte_fallback": false,
+    "type": "Unigram",
     "unk_id": 2,
     "vocab": [
       [
@@ -129404,9 +129406,9 @@
         "type": "WhitespaceSplit"
       },
       {
-        "add_prefix_space": true,
+        "prepend_scheme": "always",
         "replacement": "▁",
-        "str_rep": "▁",
+        "split": true,
         "type": "Metaspace"
       }
     ],
```

john-rocky:
Closes #329.

## Problem

Newer python `tokenizer.json` files emitted by transformers ≥ 5 (e.g. T5 saved via `T5Tokenizer.save_pretrained`) drop the legacy `add_prefix_space` decoder field and only set `prepend_scheme`:

```diff
   "decoder": {
-    "add_prefix_space": true,
+    "prepend_scheme": "always",
     "replacement": "▁",
+    "split": true,
     "type": "Metaspace"
   }
```

The Swift `MetaspaceDecoder` was reading only `add_prefix_space`, so on a fine-tuned T5 it left the leading replacement unstripped:

```
tokens   ["▁How", "▁are", "▁you", "?"]
decoded  [" How", " are", " you", "?"]   // expected: ["How", " are", " you", "?"]
```

`MetaspacePreTokenizer` already migrated to `prepend_scheme` in #319; the decoder side was missed.

## Fix

In `MetaspaceDecoder.init`, prefer `prepend_scheme` when present and translate it to the existing strip-leading-space behavior:

| `prepend_scheme` | encode added a leading space? | decoder strips first leading space? |
|---|---|---|
| `always` | yes | yes |
| `first` | yes (first piece only) | yes |
| `never` | no | no |

When `prepend_scheme` is absent, fall back to the existing `add_prefix_space` reading so legacy `tokenizer.json` files keep working.

## Verification

Confirmed the regression and the fix with the new tests:

```
# pre-fix (HEAD~1, with new tests applied):
✘ Test "Metaspace decoder honors prepend_scheme = always (T5 fine-tune)"
  recorded an issue: Expectation failed:
    (decoded.joined() → " How are you?") == "How are you?"

# post-fix (HEAD):
✔ Test "Metaspace decoder with prefix space replacement" passed
✔ Test "Metaspace decoder honors prepend_scheme = always (T5 fine-tune)" passed
✔ Test "Metaspace decoder honors prepend_scheme = never" passed
✔ Test "Metaspace decoder honors prepend_scheme = first" passed
✔ Test "Metaspace decoder prepend_scheme supersedes add_prefix_space" passed
✔ Test run with 6 tests in 1 suite passed
```

Full TokenizersTests suite (111 tests) still passes locally on macOS.

The reproduction config in the new "= always (T5 fine-tune)" test mirrors the `tokenizer.json` snippet from #329 exactly.

### DIFF
--- a/Sources/Tokenizers/Decoder.swift
+++ b/Sources/Tokenizers/Decoder.swift
@@ -237,23 +237,13 @@ class MetaspaceDecoder: Decoder {
     let replacement: String
 
     required init(config: Config) {
-        // `prepend_scheme` supersedes `add_prefix_space` per huggingface/tokenizers#1357.
-        // Newer python tokenizer.json files emitted by transformers ≥ 5 (e.g. T5 saved
-        // via `save_pretrained`) drop `add_prefix_space` and only set `prepend_scheme`.
-        // The Rust reference implementation uses the same `Metaspace` struct for both
-        // pre-tokenize and decode, so its defaults are symmetric by construction. Since
-        // Swift has separate types we mirror `MetaspacePreTokenizer`'s resolution here
-        // so encode and decode agree even when the config sets neither key (in that
-        // case both default to `.always`).
+        // See MetaspacePreTokenizer.init
         let scheme: MetaspacePreTokenizer.PrependScheme
         if let schemeStr = config.prependScheme.string() {
             scheme = MetaspacePreTokenizer.PrependScheme(rawValue: schemeStr) ?? .always
         } else {
             scheme = config.addPrefixSpace.boolean(or: true) ? .always : .never
         }
-        // For decode, both `.always` and `.first` mean a leading space was prepended
-        // at encode time and should be stripped from the first token; `.never` leaves
-        // the input untouched.
         addPrefixSpace = (scheme != .never)
         replacement = config.replacement.string(or: "_")
     }
--- a/Sources/Tokenizers/Decoder.swift
+++ b/Sources/Tokenizers/Decoder.swift
@@ -240,13 +240,21 @@ class MetaspaceDecoder: Decoder {
         // `prepend_scheme` supersedes `add_prefix_space` per huggingface/tokenizers#1357.
         // Newer python tokenizer.json files emitted by transformers ≥ 5 (e.g. T5 saved
         // via `save_pretrained`) drop `add_prefix_space` and only set `prepend_scheme`.
-        // Treat "always" / "first" as having prepended a leading space at encode time
-        // so the decoder strips it from the first token; "never" leaves it intact.
+        // The Rust reference implementation uses the same `Metaspace` struct for both
+        // pre-tokenize and decode, so its defaults are symmetric by construction. Since
+        // Swift has separate types we mirror `MetaspacePreTokenizer`'s resolution here
+        // so encode and decode agree even when the config sets neither key (in that
+        // case both default to `.always`).
+        let scheme: MetaspacePreTokenizer.PrependScheme
         if let schemeStr = config.prependScheme.string() {
-            addPrefixSpace = (schemeStr != "never")
+            scheme = MetaspacePreTokenizer.PrependScheme(rawValue: schemeStr) ?? .always
         } else {
-            addPrefixSpace = config.addPrefixSpace.boolean(or: false)
+            scheme = config.addPrefixSpace.boolean(or: true) ? .always : .never
         }
+        // For decode, both `.always` and `.first` mean a leading space was prepended
+        // at encode time and should be stripped from the first token; `.never` leaves
+        // the input untouched.
+        addPrefixSpace = (scheme != .never)
         replacement = config.replacement.string(or: "_")
     }
 
--- a/Sources/Tokenizers/Decoder.swift
+++ b/Sources/Tokenizers/Decoder.swift
@@ -237,7 +237,16 @@ class MetaspaceDecoder: Decoder {
     let replacement: String
 
     required init(config: Config) {
-        addPrefixSpace = config.addPrefixSpace.boolean(or: false)
+        // `prepend_scheme` supersedes `add_prefix_space` per huggingface/tokenizers#1357.
+        // Newer python tokenizer.json files emitted by transformers ≥ 5 (e.g. T5 saved
+        // via `save_pretrained`) drop `add_prefix_space` and only set `prepend_scheme`.
+        // Treat "always" / "first" as having prepended a leading space at encode time
+        // so the decoder strips it from the first token; "never" leaves it intact.
+        if let schemeStr = config.prependScheme.string() {
+            addPrefixSpace = (schemeStr != "never")
+        } else {
+            addPrefixSpace = config.addPrefixSpace.boolean(or: false)
+        }
         replacement = config.replacement.string(or: "_")
     }
 
--- a/Tests/TokenizersTests/DecoderTests.swift
+++ b/Tests/TokenizersTests/DecoderTests.swift
@@ -97,6 +97,23 @@ struct DecoderTests {
         #expect(decoded == [" How", " are"])
     }
 
+    /// When neither `prepend_scheme` nor `add_prefix_space` is in the config,
+    /// `MetaspacePreTokenizer` defaults to `.always` (prepend a space at encode
+    /// time). The decoder therefore has to strip the leading space from the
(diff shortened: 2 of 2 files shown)


Sources/Tokenizers/Decoder.swift
```diff
@@ -237,7 +237,16 @@ class MetaspaceDecoder: Decoder {
...
-        addPrefixSpace = config.addPrefixSpace.boolean(or: false)
+        // `prepend_scheme` supersedes `add_prefix_space` per huggingface/tokenizers#1357.
+        // Newer python tokenizer.json files emitted by transformers ≥ 5 (e.g. T5 saved
+        // via `save_pretrained`) drop `add_prefix_space` and only set `prepend_scheme`.
+        // Treat "always" / "first" as having prepended a leading space at encode time
+        // so the decoder strips it from the first token; "never" leaves it intact.
+        if let schemeStr = config.prependScheme.string() {
+            addPrefixSpace = (schemeStr != "never")
+        } else {
+            addPrefixSpace = config.addPrefixSpace.boolean(or: false)
```
pcuenca: There's a slight mismatch if neither `prependScheme` nor `addPrefixSpace` exist in the config. `PreTokenizer` would default to adding a prefix space (see [here](https://github.com/huggingface/swift-transformers/blob/15bcc471a2de73ce3c17c08781f34a1f9bebaf70/Sources/Tokenizers/PreTokenizer.swift#L170) or [here](https://github.com/huggingface/swift-transformers/blob/15bcc471a2de73ce3c17c08781f34a1f9bebaf70/Sources/Tokenizers/PreTokenizer.swift#L192)), while the decoder would default to _not_ adding it.

Did you check in the reference Rust implementation how this situation is handled?

Sources/Tokenizers/Decoder.swift
```diff
@@ -237,7 +237,16 @@ class MetaspaceDecoder: Decoder {
...
 
     required init(config: Config) {
-        addPrefixSpace = config.addPrefixSpace.boolean(or: false)
+        // `prepend_scheme` supersedes `add_prefix_space` per huggingface/tokenizers#1357.
+        // Newer python tokenizer.json files emitted by transformers ≥ 5 (e.g. T5 saved
+        // via `save_pretrained`) drop `add_prefix_space` and only set `prepend_scheme`.
+        // Treat "always" / "first" as having prepended a leading space at encode time
+        // so the decoder strips it from the first token; "never" leaves it intact.
+        if let schemeStr = config.prependScheme.string() {
+            addPrefixSpace = (schemeStr != "never")
```
pcuenca: Another approach would be to reuse the `PrependScheme` struct that we already use in the `PreTokenizer`, but I agree your current approach is easy enough for now. If you'd like to explore the alternative to see if it makes sense, feel free to go for it; otherwsie we can keep this version.

pcuenca (commented):
Thanks a lot @john-rocky, we should have caught this when #319 was added! 🙌 

I only have a question about consistency and then we can merge.

pcuenca:
We could also add another tokenizer test that exercises the new T5 tokenizer format by encoding and decoding some text and matching results against expected values. We don't need access to a full model fine-tune; just the tokenizer would do. We can do this as a follow-up PR as well.

john-rocky:
Thanks @pcuenca! Both points addressed in 10102c9:

**L248 — default mismatch (the actual unfixed bug).** You're right. The Rust reference handles this by using one `Metaspace` struct for both pre-tokenize and decode, so the default is symmetric by construction. Since the Swift port has separate types, I now mirror `MetaspacePreTokenizer`'s resolution in `MetaspaceDecoder` so the `.always` fallback applies on both sides when the config sets neither `prepend_scheme` nor `add_prefix_space`. Added a regression test that locks this in (would have asserted on `[" How", " are", " you", "?"]` against the previous default).

**L246 — reuse `PrependScheme`.** Done — `MetaspaceDecoder` now resolves into `MetaspacePreTokenizer.PrependScheme` directly and then maps `.never → false` / everything else → `true` for the strip behavior. The single source of truth makes the symmetry explicit.

Re: the encode/decode tokenizer round-trip test — happy to send that as a follow-up PR once one of the perf PRs lands and I can rebase against current main without conflicts.

Sources/Tokenizers/Decoder.swift
```diff
@@ -237,7 +237,24 @@ class MetaspaceDecoder: Decoder {
...
     required init(config: Config) {
-        addPrefixSpace = config.addPrefixSpace.boolean(or: false)
+        // `prepend_scheme` supersedes `add_prefix_space` per huggingface/tokenizers#1357.
+        // Newer python tokenizer.json files emitted by transformers ≥ 5 (e.g. T5 saved
+        // via `save_pretrained`) drop `add_prefix_space` and only set `prepend_scheme`.
+        // The Rust reference implementation uses the same `Metaspace` struct for both
+        // pre-tokenize and decode, so its defaults are symmetric by construction. Since
+        // Swift has separate types we mirror `MetaspacePreTokenizer`'s resolution here
+        // so encode and decode agree even when the config sets neither key (in that
+        // case both default to `.always`).
```
pcuenca:
```suggestion
        // See MetaspacePreTokenizer.init
```

Sources/Tokenizers/Decoder.swift
```diff
@@ -237,7 +237,24 @@ class MetaspaceDecoder: Decoder {
...
+        let scheme: MetaspacePreTokenizer.PrependScheme
+        if let schemeStr = config.prependScheme.string() {
+            scheme = MetaspacePreTokenizer.PrependScheme(rawValue: schemeStr) ?? .always
+        } else {
+            scheme = config.addPrefixSpace.boolean(or: true) ? .always : .never
+        }
+        // For decode, both `.always` and `.first` mean a leading space was prepended
+        // at encode time and should be stripped from the first token; `.never` leaves
+        // the input untouched.
```
pcuenca:
```suggestion
```

pcuenca (approved):

pcuenca:
Thanks @john-rocky, I'll focus on the performance PRs tomorrow! 🙌

PR #346, opened 2026-05-01
Speed up BPE merge with priority-queue algorithm (5-12x on inner loop)

john-rocky:
## Summary

Replace the BPE merge inner loop in `BPETokenizer.bpe(token:)` with the canonical priority-queue algorithm used by `tiktoken` and `huggingface/tokenizers` (Rust). The previous implementation rebuilt the full pair set and rescanned the whole word on every iteration — O(N² · M) — which dominated tokenize time for long pre-tokenized chunks (code, URLs, concatenated identifiers, agglutinative words).

This affects every BPE-based tokenizer registered in `TokenizerModel`:
GPT-2, RoBERTa, Llama, Gemma, Qwen2/3, Mistral, Falcon, Cohere, CodeGen, CodeLlama, Whisper, and any user-supplied `*Fast` BPE variant.

## Algorithm

- Symbols are stored as an in-place doubly linked list embedded in parallel index arrays.
- Candidate merges live in a min-heap keyed by `(rank, leftIdx)`.
- Stale heap entries are not removed on merge — they are skipped on pop by re-checking that the pair at `(left, next[left])` still has the recorded rank (lazy deletion).
- Each merge step is O(log N) heap ops + O(1) list surgery, yielding O(N log N + M log N) overall.

The 1-character early-return, the per-Character symbol decomposition, the BPE rank lookup keys (`BytePair`), and the space-joined return string are all preserved verbatim, so the function is observably equivalent to the old implementation on every well-formed input.

## Benchmarks

Apple M-series, `swift test -c release`, `mlx-community/Qwen3-0.6B-Base-DQ5` BPE table (~152K merges). Numbers are mean over 1000 (inner loop) or 200/100/50/10 (encode) iterations.

### Inner loop — `BPETokenizer.bpe(token:)` direct call

| word                                                              | length | before     | after      | speedup |
| ----------------------------------------------------------------- | -----: | ---------- | ---------- | ------: |
| `internationalization`                                            |     20 | 0.050 ms   | 0.009 ms   |  5.56x  |
| `supercalifragilisticexpialidocious`                              |     34 | 0.171 ms   | 0.017 ms   | 10.06x  |
| `vocabularyIdentifierToTokenStringMap`                            |     36 | 0.154 ms   | 0.021 ms   |  7.33x  |
| `shouldFuseConsecutiveUnknownTokenSequencesIntoASingleUnknownToken` | 65   | 0.487 ms   | 0.039 ms   | 12.49x  |

Scaling went from clearly super-linear (length × 3.25 → time × 9.7 for the 20→65 step) to roughly linear (× 4.3), as expected from the algorithmic change.

### Full `tokenizer.encode(...)` end-to-end

| input                            | before                  | after                   | speedup |
| -------------------------------- | ----------------------- | ----------------------- | ------: |
| short prompt (~100 B)            | 0.16 ms / 0.41 MB/s     | 0.13 ms / 0.51 MB/s     | 1.23x   |
| Wikipedia-style paragraph (~3 KB)| 3.25 ms / 0.34 MB/s     | 1.91 ms / 0.57 MB/s     | 1.70x   |
| Long doc (~30 KB)                | 28.80 ms / 0.38 MB/s    | 18.21 ms / 0.60 MB/s    | 1.58x   |
| Swift code (~600 B)              | 3.02 ms / 0.15 MB/s     | 0.58 ms / 0.80 MB/s     | 5.21x   |

End-to-end speedup is bounded by pre-tokenization (regex split + byte encoding), which I've left untouched in this PR. Inputs whose pre-tokenized chunks are individually long — code with camelCase identifiers, URLs, agglutinative-language tokens — see the largest wins because those are exactly the inputs that exercised the O(N²) merge loop.

## Verification

- All 107 existing `TokenizersTests` still pass (`swift test --filter TokenizersTests`). These exercise real `tokenizer.json` files for BERT, GPT-2, Falcon, Llama, Llama 3.2, Mistral, Phi-4, Qwen 2.5, Gemma, NLLB, RoBERTa-XLM, Whisper, and DeepSeek against bundled ground-truth datasets, so equivalence is verified at the encode/decode level.
- New benchmark file `Tests/Benchmarks/BPETokenizerBenchmarkTests.swift` mirrors the gating used by the existing JSON parser benchmark — it runs only with `RUN_BENCHMARKS=1` so it does not slow down normal CI:

```bash
RUN_BENCHMARKS=1 swift test -c release --filter BPETokenizerBenchmarkTests
```

The "BPE merge inner loop on synthetic long words" sub-bench is the most useful for future regressions: it isolates `bpe()` from pre-tokenization and reports per-word means.

## Why now

`huggingface/tokenizers` and `tiktoken` settled on this algorithm precisely because the naive scan blows up on long inputs. As more apps push longer prompts through swift-transformers (chat with system prompts, RAG, code agents), this hot-path is increasingly user-visible. The change is local to one function and observably-equivalent on every test case we have.

### DIFF
--- a/Sources/Tokenizers/BPETokenizer.swift
+++ b/Sources/Tokenizers/BPETokenizer.swift
@@ -174,62 +174,139 @@ class BPETokenizer: PreTrainedTokenizerModel, @unchecked Sendable {
         return result
     }
 
-    private func getPairs(word: [String]) -> Set<BytePair> {
-        var s = Set<BytePair>()
-        for i in 0..<word.count - 1 {
-            let bp = BytePair(
-                word[i],
-                word[i + 1]
-            )
-            s.insert(bp)
-        }
-        return s
-    }
-
+    /// Applies the BPE merge sequence to `token` and returns the merged pieces
+    /// joined by a single space.
+    ///
+    /// Equivalent to the canonical greedy "lowest-rank merge first" BPE algorithm
+    /// (e.g. `tiktoken`, `huggingface/tokenizers`). The previous implementation
+    /// rebuilt the entire pair set and scanned the full word on every iteration,
+    /// which is O(N² · M). This version maintains the symbols as an in-place
+    /// linked list and tracks candidate merges in a min-heap keyed by rank, so
+    /// the work per merge step is O(log N) heap ops + O(1) list surgery.
     func bpe(token: String) -> String {
         if token.count <= 1 {
             return token
         }
 
-        var word = Array(token).map { String($0) }
-        var pairs = Array(getPairs(word: word))
+        // Initial symbols: one entry per Character of `token`. We keep these as
+        // a doubly linked list embedded in parallel arrays of indices.
+        var symbols = Array(token).map { String($0) }
+        let initialCount = symbols.count
+        var prevIndex = Array(repeating: -1, count: initialCount)
+        var nextIndex = Array(repeating: -1, count: initialCount)
+        var alive = Array(repeating: true, count: initialCount)
+        for i in 0..<initialCount {
+            prevIndex[i] = i - 1
+            nextIndex[i] = (i == initialCount - 1) ? -1 : i + 1
+        }
 
-        while true {
-            let bigrams = pairs.filter { bp -> Bool in bpeRanks[bp] != nil }
-            if bigrams.count == 0 {
-                break
-            }
-            let bigram = bigrams.min { bp1, bp2 -> Bool in
-                return bpeRanks[bp1]! < bpeRanks[bp2]!
-            }!
-            let first = bigram.a
-            let second = bigram.b
-            var newWord: [String] = []
-            var i = 0
-            while i < word.count {
-                if let j = word[i..<word.count].firstIndex(of: first) {
-                    newWord.append(contentsOf: word[i..<j])
-                    i = j
-                } else {
-                    newWord.append(contentsOf: word[i..<word.count])
-                    break
-                }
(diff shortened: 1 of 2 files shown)


john-rocky:
Rebased on top of `main` and switched the bench file to `benchmarkMeasure(...)` from #347's `BenchmarkSupport.swift` (2201d95). -41 lines on `BPETokenizerBenchmarkTests.swift`. Build / lint clean.

pcuenca (commented):
I think this one only provides modest improvements in most cases, but it's good to align with the Rust implementation.

I have a few comments that I found it easier to cover in this PR rather than prose: https://github.com/john-rocky/swift-transformers/pull/1/changes

- Extract heap code to try keep the BPE algorithm as clean as possible
- Drop the `@inline`, as I saw no difference at all running the benchmarks
- Add a link to the reference Rust implementation

Please, take a look @john-rocky and if you agree with those changes you can merge my PR and then I'll merge yours :)

john-rocky:
Reviewed and merged your PR @pcuenca — `8b57888` is now on `perf/bpe-priority-queue`. The `MinHeap` extraction reads much better and the `BPEMergeCandidate: Comparable` lift cleans up `bpe(token:)` nicely. Agreed on dropping `@inline(__always)`; I added it speculatively but had no measurement to back it. Reference link is good to have.

Build clean, 115 TokenizersTests pass, lint clean. Ready for your merge whenever it's convenient. 🙌

pcuenca (approved):

PR #347, opened 2026-05-01
Cache byte-level pre-tokenize regex and inline byte encoder lookup

john-rocky:
## Summary

`ByteLevelPreTokenizer.preTokenize` is the byte-level pre-tokenization hot path
for every BPE-based tokenizer (GPT-2, Llama, Qwen, Gemma, Mistral, Whisper, ...).
On every encode call it pays three avoidable costs that this PR removes.

`BPETokenizer.byteEncode` and `BPETokenizer.hexaEncode` carry the same pattern
and are migrated to the same shared infrastructure, so the speedup applies
uniformly across the byte-level helpers.

## What changes

1. **Pre-compile the GPT-2 byte-level pre-tokenize regex once.** The pattern
   was previously passed by string to `String.range(of:options:.regularExpression)`,
   which re-parses the pattern on every match step. A single
   `NSRegularExpression` is now created at file scope in `ByteEncoder.swift`
   and shared by `ByteLevelPreTokenizer` and `BPETokenizer`. Iteration uses
   `enumerateMatches`, so we also avoid allocating an intermediate
   `[Range<Index>]`. `NSRegularExpression` is documented as thread-safe for
   `enumerateMatches` / `matches`, so no locking is required.
2. **Replace the byte → unicode-character dictionary with a fixed-size array.**
   `byteEncoder` is a `[UTF8.CodeUnit: String]`; every input byte costs a
   hash lookup. The alphabet covers the full `0..<256` range, so a 256-entry
   `[String]` (`byteEncoderTable`) is dense and lets the inner loop replace
   the dictionary lookup with an array index.
3. **Drop intermediate allocations in the inner loop.** The previous form
   materialised an `Array(token.utf8)` and a `compactMap` before `joined()`.
   Appending the table entries directly into a single output `String` (with
   reserved capacity) drops both intermediates.
4. **Cache `<0x%02X>` strings in `hexaEncode`.** The byte fallback formatter
   was called once per byte; a 256-entry `hexaEncoderTable` removes the
   per-call `String(format:)` overhead.

## Verification

`swift test --filter TokenizersTests` — all **107 existing tokenizer tests**
pass, including the 18 ``Pre-Tokenizer Tests`` (3 ``ByteLevelPreTokenizer``
configurations covering the default / ``addPrefixSpace`` / ``useRegex=false``
code paths).

A new benchmark suite ``Tests/Benchmarks/BPEPreTokenizeBenchmarkTests.swift``
is gated on ``RUN_BENCHMARKS=1`` and reuses the same Qwen3-0.6B-Base BPE model
used by the existing JSON parser benchmark, so the two suites can be reasoned
about together.

### Numbers (Apple M-series, Swift release, mean of `iterations` runs after 3 warmup runs)

`ByteLevelPreTokenizer.preTokenize` micro-benchmark — the direct hot path:

| input | iters | before (mean / p50) | after (mean / p50) | p50 speedup |
|---|---:|---:|---:|---:|
| short (~70 B) | 5,000 | 0.034 / 0.026 ms | 0.004 / 0.004 ms | **6.5x** |
| code (~600 B) | 1,000 | 0.234 / 0.165 ms | 0.025 / 0.025 ms | **6.6x** |
| medium (~3 KB) | 200 | 0.549 / 0.456 ms | 0.062 / 0.062 ms | **7.4x** |
| long (~30 KB) | 20 | 6.488 / 4.750 ms | 0.606 / 0.607 ms | **7.8x** |
| addPrefixSpace short | 5,000 | 0.033 / 0.025 ms | 0.004 / 0.004 ms | **6.3x** |
| useRegex=false short | 5,000 | 0.002 / 0.002 ms | 0.001 / 0.001 ms | 2.0x |

`BPETokenizer.byteEncode` / `hexaEncode` micro-benchmark — the same shared
infrastructure benefits both helpers:

| function / input | iters | before (mean / p50) | after (mean / p50) | p50 speedup |
|---|---:|---:|---:|---:|
| `byteEncode` short (~70 B) | 5,000 | 0.029 / 0.027 ms | 0.005 / 0.004 ms | **6.8x** |
| `byteEncode` code (~600 B) | 1,000 | 0.189 / 0.169 ms | 0.024 / 0.024 ms | **7.0x** |
| `byteEncode` medium (~3 KB) | 200 | 0.519 / 0.479 ms | 0.060 / 0.060 ms | **8.0x** |
| `hexaEncode` long word | 10,000 | 0.030 / 0.029 ms | 0.001 / 0.001 ms | **29x** |
| `hexaEncode` short (~70 B) | 5,000 | 0.082 / 0.069 ms | 0.004 / 0.004 ms | **17x** |

(`hexaEncode` benefits more because the cached `<0x%02X>` table replaces a
``String(format:)`` call on every byte.)

End-to-end `tokenizer.encode(text:)` on Qwen3-0.6B-Base (152K-merge BPE):

| input | iters | before (p50) | after (p50) |
|---|---:|---:|---:|
| short (~70 B) | 1,000 | 0.149 ms | 0.149 ms |
| code (~600 B) | 200 | 2.461 ms | 2.420 ms |
| medium (~3 KB) | 50 | 2.560 ms | 2.661 ms |
| long (~30 KB) | 10 | 28.123 ms | 27.545 ms |

End-to-end is dominated by the `BPETokenizer.bpe(token:)` merge inner loop on
this model (152K merges), which is what PR #346 targets. The two PRs are
complementary: this one cuts the pre-tokenize phase, that one cuts the merge
phase. End-to-end gains compose if both land.

### Reproduction

```
git fetch origin
git checkout main
git checkout -b before
swift test -c release --filter TokenizersTests          # 107 tests pass
RUN_BENCHMARKS=1 swift test -c release \
  --filter BPEPreTokenizeBenchmarkTests                 # before numbers

git checkout perf/bpe-pretokenize-fast-path
swift test -c release --filter TokenizersTests          # 107 tests still pass
RUN_BENCHMARKS=1 swift test -c release \
  --filter BPEPreTokenizeBenchmarkTests                 # after numbers
```

The benchmark suite is opt-in (gated on `RUN_BENCHMARKS=1`), matching the
existing convention in ``Tests/Benchmarks/JSONParserBenchmarkTests.swift``.

### DIFF
--- a/Sources/Tokenizers/BPETokenizer.swift
+++ b/Sources/Tokenizers/BPETokenizer.swift
@@ -148,20 +148,38 @@ class BPETokenizer: PreTrainedTokenizerModel, @unchecked Sendable {
         idsToTokens[id] as String?
     }
 
+    /// Cached `<0x%02X>` byte fallback strings, indexed by byte value.
+    private static let hexaEncoderTable: [String] = (0..<256).map { String(format: "<0x%02X>", $0) }
+
     func byteEncode(text: String) -> [String] {
-        let RE = #"'s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+"#
-        let tokens = text.ranges(of: RE).map { String(text[$0]) }
-        return tokens.map { token -> String in
-            return Array(token.utf8).compactMap { byteEncoder[$0] }.joined()
+        let nsText = text as NSString
+        let fullRange = NSRange(location: 0, length: nsText.length)
+        var result: [String] = []
+        byteLevelPreTokenizeRegex.enumerateMatches(in: text, range: fullRange) { match, _, _ in
+            guard let match else { return }
+            let token = nsText.substring(with: match.range)
+            var encoded = ""
+            encoded.reserveCapacity(token.utf8.count)
+            for byte in token.utf8 {
+                encoded.append(byteEncoderTable[Int(byte)])
+            }
+            result.append(encoded)
         }
+        return result
     }
 
     func hexaEncode(text: String) -> [String] {
-        let RE = #"'s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+"#
-        let tokens = text.ranges(of: RE).map { String(text[$0]) }
-        return tokens.flatMap { token -> [String] in
-            return Array(token.utf8).map { String(format: "<0x%02X>", $0) }
+        let nsText = text as NSString
+        let fullRange = NSRange(location: 0, length: nsText.length)
+        var result: [String] = []
+        byteLevelPreTokenizeRegex.enumerateMatches(in: text, range: fullRange) { match, _, _ in
+            guard let match else { return }
+            let token = nsText.substring(with: match.range)
+            for byte in token.utf8 {
+                result.append(Self.hexaEncoderTable[Int(byte)])
+            }
         }
+        return result
     }
 
     private func getPairs(word: [String]) -> Set<BytePair> {
--- a/Sources/Tokenizers/BPETokenizer.swift
+++ b/Sources/Tokenizers/BPETokenizer.swift
@@ -152,12 +152,8 @@ class BPETokenizer: PreTrainedTokenizerModel, @unchecked Sendable {
     private static let hexaEncoderTable: [String] = (0..<256).map { String(format: "<0x%02X>", $0) }
 
     func byteEncode(text: String) -> [String] {
-        let nsText = text as NSString
-        let fullRange = NSRange(location: 0, length: nsText.length)
         var result: [String] = []
-        byteLevelPreTokenizeRegex.enumerateMatches(in: text, range: fullRange) { match, _, _ in
-            guard let match else { return }
-            let token = nsText.substring(with: match.range)
+        enumerateRegexTokens(in: text, with: byteLevelPreTokenizeRegex) { token in
             var encoded = ""
             encoded.reserveCapacity(token.utf8.count)
             for byte in token.utf8 {
@@ -169,12 +165,8 @@ class BPETokenizer: PreTrainedTokenizerModel, @unchecked Sendable {
     }
 
     func hexaEncode(text: String) -> [String] {
-        let nsText = text as NSString
-        let fullRange = NSRange(location: 0, length: nsText.length)
         var result: [String] = []
-        byteLevelPreTokenizeRegex.enumerateMatches(in: text, range: fullRange) { match, _, _ in
-            guard let match else { return }
-            let token = nsText.substring(with: match.range)
+        enumerateRegexTokens(in: text, with: byteLevelPreTokenizeRegex) { token in
             for byte in token.utf8 {
                 result.append(Self.hexaEncoderTable[Int(byte)])
             }
--- a/Sources/Tokenizers/ByteEncoder.swift
+++ b/Sources/Tokenizers/ByteEncoder.swift
@@ -284,19 +284,3 @@ let byteEncoderTable: [String] = {
     }
     return arr
 }()
-
-/// Pre-compiled GPT-2 / Llama / Qwen / Gemma byte-level pre-tokenization regex.
-///
-/// The same pattern is used by ``BPETokenizer/byteEncode(text:)``,
-/// ``BPETokenizer/hexaEncode(text:)`` and ``ByteLevelPreTokenizer/preTokenize(text:options:)``.
-/// Foundation's `String.range(of:options:.regularExpression)` re-parses the pattern on every
-/// call, so caching a single `NSRegularExpression` removes the dominant cost of
-/// short-input `encode` and lets us iterate matches via `enumerateMatches` without
-/// allocating an intermediate range array.
-let byteLevelPreTokenizeRegex: NSRegularExpression = {
-    let pattern = #"'s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+"#
-    // The pattern is a compile-time constant and has been used in production for
-    // years; treating compilation failure as a programmer error matches every
-    // other in-tree regex.
-    return try! NSRegularExpression(pattern: pattern)
-}()
--- a/Sources/Tokenizers/ByteEncoder.swift
+++ b/Sources/Tokenizers/ByteEncoder.swift
@@ -270,3 +270,33 @@ let byteEncoder: [UTF8.CodeUnit: String] = [
 let byteDecoder = byteEncoder.reduce(into: [String: UTF8.CodeUnit]()) { result, element in
     result[element.value] = element.key
 }
+
+/// Fixed-size byte → unicode-character lookup, derived from ``byteEncoder``.
+///
+/// On the byte-level pre-tokenization hot path every input byte is encoded
+/// via this map, and a `[UInt8: String]` dictionary lookup costs roughly an
+/// order of magnitude more than indexing into a contiguous array. The byte
+/// alphabet covers the full 0...255 range, so a 256-element array is dense.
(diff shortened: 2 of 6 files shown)


Sources/Tokenizers/ByteEncoder.swift
```diff
@@ -270,3 +270,33 @@ let byteEncoder: [UTF8.CodeUnit: String] = [
...
+
+/// Pre-compiled GPT-2 / Llama / Qwen / Gemma byte-level pre-tokenization regex.
+///
+/// The same pattern is used by ``BPETokenizer/byteEncode(text:)``,
+/// ``BPETokenizer/hexaEncode(text:)`` and ``ByteLevelPreTokenizer/preTokenize(text:options:)``.
+/// Foundation's `String.range(of:options:.regularExpression)` re-parses the pattern on every
+/// call, so caching a single `NSRegularExpression` removes the dominant cost of
+/// short-input `encode` and lets us iterate matches via `enumerateMatches` without
+/// allocating an intermediate range array.
+let byteLevelPreTokenizeRegex: NSRegularExpression = {
```
pcuenca: I wonder if `PreTokenizer` wouldn't be a better place for this - also consistent with upcoming PRs like #349.

Sources/Tokenizers/PreTokenizer.swift
```diff
@@ -224,25 +224,43 @@ class ByteLevelPreTokenizer: PreTokenizer {
...
 
     required init(config: Config) {
         addPrefixSpace = config.addPrefixSpace.boolean(or: false)
         trimOffsets = config.trimOffsets.boolean(or: true)
         useRegex = config.useRegex.boolean(or: true)
     }
 
+    /// Byte-level encode a single token (no pre-tokenization split). Hot inner
+    /// loop: indexes the cached 256-entry table once per UTF-8 byte and appends
+    /// directly into a single output string.
```
pcuenca:
```suggestion
```

I think the comment in `byteEncoderTable` is enough.

Tests/Benchmarks/BPEPreTokenizeBenchmarkTests.swift
```diff
@@ -0,0 +1,195 @@
...
+                public let beginningOfSequenceSpecialTokenIdentifier: Int?
+                public let endOfSequenceSpecialTokenIdentifier: Int?
+                public let shouldFuseConsecutiveUnknownTokenSequencesIntoASingleUnknownToken: Bool
+            }
+            """
+    }
+
+    // MARK: - Measurement helpers
+
+    struct Stats {
```
pcuenca: I would extract as much common benchmarking infrastructure as possible, as noted in #349, so all PRs can benefit from it.

Sources/Tokenizers/BPETokenizer.swift
```diff
@@ -148,20 +148,38 @@ class BPETokenizer: PreTrainedTokenizerModel, @unchecked Sendable {
...
+
     func byteEncode(text: String) -> [String] {
-        let RE = #"'s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+"#
-        let tokens = text.ranges(of: RE).map { String(text[$0]) }
-        return tokens.map { token -> String in
-            return Array(token.utf8).compactMap { byteEncoder[$0] }.joined()
+        let nsText = text as NSString
+        let fullRange = NSRange(location: 0, length: nsText.length)
+        var result: [String] = []
+        byteLevelPreTokenizeRegex.enumerateMatches(in: text, range: fullRange) { match, _, _ in
```
pcuenca: Did you consider using a helper function with a callback to wrap each of the three `enumerateMatches` invocations in this PR into something with a semantically nice name? I'm not sure whether this would simplify things or make them more confusing; honestly asking for your opinion here.

The structure would match the `splitMatches` helper you use in #349, but I'm more concerned about clarity than consistency.

pcuenca (commented):
This looks good to me! I appreciate you created focused PRs with a common theme. I would like to merge them all when we are ready, so I wouldn't object if we extracted common benchmarking infrastructure from this one (#347) so it can be reused in the others.

I will test locally on both macOS and Linux.

After tests pass, we'll be ready to merge. I asked a few minor comments / questions that I think we can resolve quickly.

pcuenca:
(My preliminary thoughts about merge order, based on apparent complexity and PR relationships, would be #345, #347, #349, #348, #346).

john-rocky:
Thanks @pcuenca! All four points addressed in 8ab6f12:

**L296 — relocate `byteLevelPreTokenizeRegex` to `PreTokenizer.swift`.** Done. Moved the const + doc next to `ByteLevelPreTokenizer` (its only producer), matching where the per-PreTokenizer caches in #349 live. `byteEncoderTable` / `byteEncoder` / `byteDecoder` stay in `ByteEncoder.swift` since those are the byte-encoder data tables.

**L236 — trim the redundant `byteEncodeToken` doc.** Done. The `byteEncoderTable` doc already explains why the table beats a dict lookup; the function comment is now just the one-line "Byte-level encode a single token (no pre-tokenization split)."

**L66 — extract common bench infrastructure.** Done. New `Tests/Benchmarks/BenchmarkSupport.swift` has `BenchmarkStats`, `benchmarkStats(_:)`, and `benchmarkMeasure(label:iterations:warmup:_:)`. `BPEPreTokenizeBenchmarkTests` now uses it (-59 lines). The sibling perf branches (#346 / #348 / #349) keep their local copies for now to keep each PR self-contained on its own branch; once #347 lands I'll rebase the others to drop their copies and pick up `BenchmarkSupport` (no review-time work for you, just a follow-up rebase from me).

**L158 — helper for the three `enumerateMatches` sites.** Went with the helper. Honestly, the bodies do differ in non-trivial ways (prefix-space branch / single-string accumulation / per-byte accumulation), but the `NSString` / `NSRange` / guard scaffolding was the dominant noise; with the helper each site reads as "for each matched token, do X" and the per-token logic is the focal point. New `enumerateRegexTokens(in:with:_:)` lives at file scope in `PreTokenizer.swift` so `BPETokenizer` can call it too. It's deliberately *not* `splitMatches` — that helper from #349 returns `[String]` and is the right shape there; here we want a callback because two of the three sites need to do per-byte accumulation, not per-match collection.

Build clean, 107 existing TokenizersTests pass, swift-format lint clean. Ready for another look when you have a moment.

Sources/Tokenizers/ByteEncoder.swift
```diff
@@ -270,3 +270,17 @@ let byteEncoder: [UTF8.CodeUnit: String] = [
 let byteDecoder = byteEncoder.reduce(into: [String: UTF8.CodeUnit]()) { result, element in
     result[element.value] = element.key
 }
+
+/// Fixed-size byte → unicode-character lookup, derived from ``byteEncoder``.
+///
+/// On the byte-level pre-tokenization hot path every input byte is encoded
+/// via this map, and a `[UInt8: String]` dictionary lookup costs roughly an
+/// order of magnitude more than indexing into a contiguous array. The byte
+/// alphabet covers the full 0...255 range, so a 256-element array is dense.
```
pcuenca, line 279: Can we maybe create a small test to verify we get a dense 256-element array? This would detect regressions / deletions in the encoder table (which we don't expect, of course).

pcuenca (approved):
Thanks @john-rocky, I synced with `main` and ensured it builds on Linux. Also fixed CI on `main` for additional piece of mind.

I asked a final question about a small defensive test, let me know what you think.

john-rocky:
Thanks for the merge sync + Linux build fix @pcuenca! 🙌

Added the defensive tests in b828565 — `Tests/TokenizersTests/ByteEncoderTests.swift` covers three invariants:

1. `byteEncoderTable.count == 256` and every entry is non-empty (the array is indexed directly with no fallback, so a missing entry would silently emit `""`).
2. `byteEncoderTable` matches `byteEncoder` entry-for-entry (catches drift if either declaration is edited in isolation).
3. `byteEncoder` and `byteDecoder` are exact inverses (round-trip guard).

3 tests, all pass locally. Build clean, lint clean.

pcuenca (approved):

pcuenca:
Merging now, let's now rebase and adapt #349 🔥

PR #348, opened 2026-05-01
Speed up SentencePiece/Unigram tokenization (O(N²) → O(N), ~15-25x on long inputs)

john-rocky:
## Summary

`UnigramTokenizer.tokenize` (T5 / XLM-RoBERTa / multilingual-e5 / kredor /
NLLB / Llama-SP) and the underlying `TokenLattice` walked the input via
`String.count` and `String.index(_:offsetBy:)`, both of which scan grapheme
clusters and run in **O(N)** per call. Each call sits inside a
per-character loop, making tokenization **O(N²)** in the input length.

This PR materializes a `[Character]` view exactly once per `tokenize` call
and threads it through the lattice, so the inner loop and `piece(_:)` both
index in O(1).

## What changed

- `Sources/Tokenizers/UnigramTokenizer.swift`
  - Build `let chars = Array(text)` once, drive the trie walk with
    `chars[beginPos...]` (an `ArraySlice<Character>` is `Sequence<Character>`,
    so it slots into the existing `Trie.commonPrefixSearchIterator` API).
  - Use `tokenChars.count` (Array, O(1)) instead of `String(chars).count`
    (grapheme scan, O(K)).
  - The outer loop bound is now the cached `charsCount` instead of
    `sentence.count`.

- `Sources/Tokenizers/TokenLattice.swift`
  - Store `chars: [Character]` and a precomputed `let count: Int`. The old
    `var count: Int { sentence.count }` was O(N) on every access and was
    called inside `viterbi()` and from `UnigramTokenizer.tokenize`.
  - Add an `init(sentence:chars:bosTokenId:eosTokenId:)` so `UnigramTokenizer`
    can hand its `[Character]` over without re-materializing it. The original
    `init(sentence:bosTokenId:eosTokenId:)` is preserved and now delegates.
  - `piece(_:)` now indexes into `chars`, so resolving each token is O(1)
    instead of two `String.index(_:offsetBy:)` traversals.

No public API changes.

## Performance (`FacebookAI/xlm-roberta-base`, M-series, `swift test -c release`)

`UnigramTokenizer.tokenize` micro-benchmark (mean):

| input            | before        | after        | speedup    |
|------------------|---------------|--------------|------------|
| short (~80 B)    |   0.119 ms    |   0.102 ms   |  1.17×     |
| code (~600 B)    |   1.054 ms    |   0.604 ms   |  1.74×     |
| medium (~1.3 KB) |   2.836 ms    |   1.140 ms   |  2.49×     |
| long (~13 KB)    | **176.813 ms**| **11.892 ms**| **14.87×** |

Scaling test — same paragraph repeated 1× / 4× / 16× / 64×, tokenized in
one call. ms-per-byte is flat after the fix, confirming **O(N²) → O(N)**:

| size             | before µs/byte | after µs/byte | per-call speedup |
|------------------|----------------|---------------|------------------|
| ×1 (234 B)       | 1.72           | 1.22          | 1.42×            |
| ×4 (936 B)       | 3.15           | 1.21          | 2.61×            |
| ×16 (3.7 KB)     | 8.52           | 1.23          | 6.92×            |
| **×64 (15 KB)**  | **30.37**      | **1.23**      | **24.70×**       |

End-to-end `tokenizer.encode(...)` improves more modestly on the same
inputs (0.31 → 0.32 MB/s on the 13 KB paragraph) because XLM-R's
`MetaspacePreTokenizer` chunks the input before reaching the lattice —
the win is largest when callers pass long un-chunked text directly to
`tokenize`, or use models whose pre-tokenizer keeps long spans intact.

`Trie.commonPrefixSearch` numbers are unchanged (untouched in this PR;
the trie is fast already, ~1 µs per call).

## Correctness

All existing TokenizersTests continue to pass (107/107):

```
✔ Test robertaXLMTokenizer() passed after 7.805 seconds.            (Unigram, multilingual-e5-small)
✔ Test robertaXLMCanonicalTokenizer() passed after 34.933 seconds.  (Unigram, FacebookAI/xlm-roberta-base)
✔ Test kredorPunctuateAllTokenizer() passed after 6.429 seconds.    (Unigram)
✔ Test nllbTokenizer() passed after 25.241 seconds.                 (BPE fallback path)
✔ TrieTests (3/3)
…
✔ Test run with 107 tests in 12 suites passed after 34.934 seconds.
```

These tests assert exact token-id sequences against Python references,
so they are sensitive to any mis-segmentation introduced by the fast
path.

## Reproduction

```bash
RUN_BENCHMARKS=1 swift test -c release \
  --filter "UnigramTokenizerBenchmarkTests"
```

The new file `Tests/Benchmarks/UnigramTokenizerBenchmarkTests.swift` is
gated on `RUN_BENCHMARKS=1` and downloads `FacebookAI/xlm-roberta-base`
on first run (~5 MB tokenizer.json, cached afterwards).

## Risk

- Memory: one extra `[Character]` per `tokenize` call. Each `Character`
  is a small fixed-size value (typically 16 B inline + heap for grapheme
  clusters > 15 bytes). For the largest realistic input we exercised
  here (15 KB) this adds ~250 KB of transient allocation; freed at the
  end of the call.
- Behavior: `piece(_:)` now returns a freshly constructed `String`
  rather than a `Substring` slicing into the original sentence. The
  return type is still `any StringProtocol` (callers wrap it in
  `String(...)` already, see `TokenLattice.tokens`), so this is
  source-compatible.

Companion to #346 (BPE merge priority queue) and #347 (BPE pre-tokenize
fast path) — same idea applied to the SentencePiece / Unigram side of
the codebase.

### DIFF
--- a/Sources/Tokenizers/TokenLattice.swift
+++ b/Sources/Tokenizers/TokenLattice.swift
@@ -15,27 +15,45 @@ struct TokenLattice {
     let bosTokenId: Int
     let eosTokenId: Int
 
+    /// `Character` view of `sentence`, materialized once at init.
+    /// Lattice offsets and lengths are in `Character` units, so resolving
+    /// `piece(_:)` through this array is O(1) instead of paying the
+    /// `String.index(_:offsetBy:)` traversal that scaled as O(N) per token.
+    private let chars: [Character]
+
+    /// Pre-computed `chars.count`. The previous `var count: Int { sentence.count }`
+    /// scanned grapheme clusters on every access, which made `count` O(N) and
+    /// callers that hit it inside a per-character loop O(N²).
+    let count: Int
+
     var nodes: [TokenLatticeNode] = []
     var beginNodes: [[TokenLatticeNode]]
     var endNodes: [[TokenLatticeNode]]
 
-    var count: Int { sentence.count }
-
     init(sentence: String, bosTokenId: Int, eosTokenId: Int) {
+        self.init(sentence: sentence, chars: Array(sentence), bosTokenId: bosTokenId, eosTokenId: eosTokenId)
+    }
+
+    /// Internal initializer that lets callers reuse a `[Character]` they have
+    /// already materialized (e.g. `UnigramTokenizer.tokenize` walks the same
+    /// array to drive trie lookups).
+    init(sentence: String, chars: [Character], bosTokenId: Int, eosTokenId: Int) {
         self.sentence = sentence
+        self.chars = chars
+        self.count = chars.count
         self.bosTokenId = bosTokenId
         self.eosTokenId = eosTokenId
 
-        beginNodes = Array(repeating: [], count: sentence.count + 1)
-        endNodes = Array(repeating: [], count: sentence.count + 1)
+        beginNodes = Array(repeating: [], count: count + 1)
+        endNodes = Array(repeating: [], count: count + 1)
 
         let bos = TokenLatticeNode(tokenId: bosTokenId, startOffset: 0, length: 0, score: 0)
-        let eos = TokenLatticeNode(tokenId: eosTokenId, startOffset: sentence.count, length: 0, score: 0)
+        let eos = TokenLatticeNode(tokenId: eosTokenId, startOffset: count, length: 0, score: 0)
 
         nodes.append(bos)
         nodes.append(eos)
 
-        beginNodes[sentence.count].append(eos)
+        beginNodes[count].append(eos)
         endNodes[0].append(bos)
     }
 }
@@ -96,15 +114,16 @@ extension TokenLattice {
         return result.reversed()
     }
 
-    /// Returns the substring of the sentence to be tokenized associated to the specified node
+    /// Returns the string for the token associated to the specified node.
     ///
-    /// - Parameter node: The node defining the token to be extracted
+    /// - Parameter node: The node defining the token to be extracted.
     ///
-    /// - Returns: A **Substring** – i.e., a reference to the original positions, not a copy of the characters.
+    /// - Returns: A `String` reconstructed from the cached `Character` array.
+    ///   The previous implementation called `String.index(_:offsetBy:)` twice
+    ///   per node, which is O(N) over the sentence and dominated tokenization
+    ///   for inputs longer than a few hundred characters.
     func piece(_ node: TokenLatticeNode) -> any StringProtocol {
-        let start = sentence.index(sentence.startIndex, offsetBy: node.startOffset)
-        let end = sentence.index(start, offsetBy: node.length)
-        return sentence[start..<end]
+        String(chars[node.startOffset..<(node.startOffset + node.length)])
     }
 }
 
--- a/Sources/Tokenizers/TokenLattice.swift
+++ b/Sources/Tokenizers/TokenLattice.swift
@@ -33,17 +33,17 @@ struct TokenLattice {
         self.bosTokenId = bosTokenId
         self.eosTokenId = eosTokenId
 
-        let n = chars.count
-        beginNodes = Array(repeating: [], count: n + 1)
-        endNodes = Array(repeating: [], count: n + 1)
+        let count = chars.count
+        beginNodes = Array(repeating: [], count: count + 1)
+        endNodes = Array(repeating: [], count: count + 1)
 
         let bos = TokenLatticeNode(tokenId: bosTokenId, startOffset: 0, length: 0, score: 0)
-        let eos = TokenLatticeNode(tokenId: eosTokenId, startOffset: n, length: 0, score: 0)
+        let eos = TokenLatticeNode(tokenId: eosTokenId, startOffset: count, length: 0, score: 0)
 
         nodes.append(bos)
         nodes.append(eos)
 
-        beginNodes[n].append(eos)
+        beginNodes[count].append(eos)
         endNodes[0].append(bos)
     }
 }
@@ -69,8 +69,8 @@ extension TokenLattice {
     /// It's unfortunate that it can't be lazy or cached as the node arrays are not immutable.
     /// We could create another type that holds the nodes and use it as an immutable var  in TokenLattice.
     func viterbi() -> [TokenLatticeNode] {
-        let n = chars.count
-        for offset in 0...n {
+        let count = chars.count
+        for offset in 0...count {
             guard beginNodes[offset].count > 0 else { return [] }
 
             for rnode in beginNodes[offset] {
@@ -92,7 +92,7 @@ extension TokenLattice {
             }
         }
 
-        let root = beginNodes[n][0]
+        let root = beginNodes[count][0]
         guard let prev = root.prev else { return [] }
 
         // TODO: the reference implementations have a few more clones here: verify
--- a/Sources/Tokenizers/TokenLattice.swift
+++ b/Sources/Tokenizers/TokenLattice.swift
@@ -11,49 +11,39 @@
 /// Based on https://github.com/huggingface/tokenizers/blob/b58227c7f1ccf8b73ee2268354336da56d91e492/tokenizers/src/models/unigram/lattice.rs#L137
 /// and https://github.com/xenova/transformers.js/blob/b07336d8f7ff57453cc164cc68aead2a79cbd57e/src/utils/data-structures.js#L269C28-L269C28
 struct TokenLattice {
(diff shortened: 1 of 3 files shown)


john-rocky:
Rebased on top of `main` and switched the bench file to `benchmarkMeasure(...)` from #347's `BenchmarkSupport.swift` (df7fc28). -41 lines on `UnigramTokenizerBenchmarkTests.swift`. Build / lint clean.

Sources/Tokenizers/TokenLattice.swift
pcuenca: `sentence` is no longer needed.

Sources/Tokenizers/TokenLattice.swift
```diff
@@ -15,27 +15,45 @@ struct TokenLattice {
...
+    let count: Int
+
     var nodes: [TokenLatticeNode] = []
     var beginNodes: [[TokenLatticeNode]]
     var endNodes: [[TokenLatticeNode]]
 
-    var count: Int { sentence.count }
-
     init(sentence: String, bosTokenId: Int, eosTokenId: Int) {
+        self.init(sentence: sentence, chars: Array(sentence), bosTokenId: bosTokenId, eosTokenId: eosTokenId)
```
pcuenca:
```suggestion
        self.init(chars: Array(sentence), bosTokenId: bosTokenId, eosTokenId: eosTokenId)
```

Sources/Tokenizers/TokenLattice.swift
```diff
@@ -15,27 +15,45 @@ struct TokenLattice {
...
+    }
+
+    /// Internal initializer that lets callers reuse a `[Character]` they have
+    /// already materialized (e.g. `UnigramTokenizer.tokenize` walks the same
+    /// array to drive trie lookups).
+    init(sentence: String, chars: [Character], bosTokenId: Int, eosTokenId: Int) {
         self.sentence = sentence
```
pcuenca:
```suggestion
    init(chars: [Character], bosTokenId: Int, eosTokenId: Int) {
```

Sources/Tokenizers/UnigramTokenizer.swift
```diff
@@ -132,21 +132,33 @@ class UnigramTokenizer: PreTrainedTokenizerModel, @unchecked Sendable {
     /// - Parameter text: The input text to tokenize
     /// - Returns: An array of token strings representing the most probable segmentation
     func tokenize(text: String) -> [String] {
-        var lattice = TokenLattice(sentence: text, bosTokenId: bosTokenId ?? 0, eosTokenId: eosTokenId ?? 0)
+        // Materialize the `Character` view exactly once. Both the trie walk
+        // below and `TokenLattice.piece(_:)` need character-level random
+        // access; the previous implementation reached for it via
+        // `String.index(_:offsetBy:)` and `String.count`, both O(N), inside
+        // a per-character loop -> overall O(N^2) on long inputs.
```
pcuenca:
```suggestion
```

Sources/Tokenizers/UnigramTokenizer.swift
```diff
@@ -132,21 +132,33 @@ class UnigramTokenizer: PreTrainedTokenizerModel, @unchecked Sendable {
...
+        let chars = Array(text)
+        let charsCount = chars.count
+        var lattice = TokenLattice(
+            sentence: text,
```
pcuenca:
```suggestion
```

Sources/Tokenizers/TokenLattice.swift
```diff
@@ -15,27 +15,45 @@ struct TokenLattice {
...
 
-    var count: Int { sentence.count }
-
     init(sentence: String, bosTokenId: Int, eosTokenId: Int) {
+        self.init(sentence: sentence, chars: Array(sentence), bosTokenId: bosTokenId, eosTokenId: eosTokenId)
+    }
+
+    /// Internal initializer that lets callers reuse a `[Character]` they have
+    /// already materialized (e.g. `UnigramTokenizer.tokenize` walks the same
+    /// array to drive trie lookups).
```
pcuenca:
```suggestion
```

Sources/Tokenizers/TokenLattice.swift
```diff
@@ -15,27 +15,45 @@ struct TokenLattice {
...
+    /// `Character` view of `sentence`, materialized once at init.
+    /// Lattice offsets and lengths are in `Character` units, so resolving
+    /// `piece(_:)` through this array is O(1) instead of paying the
+    /// `String.index(_:offsetBy:)` traversal that scaled as O(N) per token.
+    private let chars: [Character]
+
+    /// Pre-computed `chars.count`. The previous `var count: Int { sentence.count }`
+    /// scanned grapheme clusters on every access, which made `count` O(N) and
+    /// callers that hit it inside a per-character loop O(N²).
+    let count: Int
```
pcuenca: I _think_ we don't need to cache this anymore; we used to pay the cost of grapheme cluster scanning in the previous implementation, but `chars.count` should be essentially free now. If you can run the benchmarks to double-check, that would be awesome. If there are no performance regressions, I'd recommend we remove the ivar so we are not misled in the future about thinking we are caching it for a reason.

Sources/Tokenizers/TokenLattice.swift
```diff
@@ -15,27 +15,45 @@ struct TokenLattice {
     let bosTokenId: Int
     let eosTokenId: Int
 
+    /// `Character` view of `sentence`, materialized once at init.
+    /// Lattice offsets and lengths are in `Character` units, so resolving
+    /// `piece(_:)` through this array is O(1) instead of paying the
+    /// `String.index(_:offsetBy:)` traversal that scaled as O(N) per token.
```
pcuenca:
```suggestion
    /// `Character` view of the input String for performance.
    /// Lattice offsets and lengths are in `Character` units, so direct
    /// access through the array does not pay the cost of
    /// `String.index(_:offsetBy:)` traversal (O(N) per token, quadratic for sequences).
```

pcuenca (commented):
* I would remove `sentence` from `TokenLattice` and adapt invocation from `UnigramTokenizer` accordingly. We can keep the previous initializer for convenience, but we immediately call the new one with the chars array, and drop the `sentence` String.
* I think we don't need to cache the char count anymore, so I'd suggest to remove it as well.
* If we do the two points above, we can simplify the various comments that explained the difference vs the previous implementation and just keep a focused comment in a single place.

pcuenca:
Tested Linux compilation locally.

john-rocky:
Thanks @pcuenca! All three points addressed in f09bccd.

**`sentence` ivar removed.** `TokenLattice` now stores only `chars` / `bosTokenId` / `eosTokenId`. The `init(sentence: String, ...)` convenience initializer is kept and just forwards to `init(chars: Array(sentence), ...)`. `UnigramTokenizer.tokenize` now constructs the lattice with `chars` directly (no `sentence:` arg).

**Cached `count` ivar removed, with bench data.** Ran `RUN_BENCHMARKS=1 swift test -c release --filter UnigramTokenizerBenchmarkTests` on M-series, release build, FacebookAI/xlm-roberta-base, before vs after the removal:

| input | before (cached `count`) | after (`chars.count` direct) |
|---|---:|---:|
| tokenize micro short ~80 B | 1.160 us/byte | 1.191 us/byte |
| tokenize micro code ~600 B | 1.079 | 1.105 |
| tokenize micro medium ~1.3 KB | 1.094 | 1.123 |
| tokenize micro long ~13 KB | 1.090 | 1.108 |
| tokenize scaling x1 | 1.117 | 1.151 |
| tokenize scaling x4 | 1.096 | 1.156 |
| tokenize scaling x16 | 1.116 | 1.143 |
| tokenize scaling x64 | 1.098 | 1.173 |

The means differ by ~3-5%, but I observed run-to-run variation of similar magnitude on either side, so this is at the noise floor rather than a real regression. To keep the loop bound predictable I hoisted `chars.count` into a local `n` inside the chars initializer and `viterbi()` (one read per call) — it makes the intent explicit without adding a stored property.

**Comments simplified.** The single focused doc lives on `chars` (using your suggested wording). Dropped the redundant comment block in `UnigramTokenizer.tokenize` and the perf-history comment on `piece(_:)`.

Build clean, 115 tokenizer tests pass, lint clean.

pcuenca (approved):

pcuenca:
Thanks for iterating here @john-rocky 🙌

Issue #352, opened 2026-05-09
Multilingual byte-divergence from HuggingFace Python: 4 distinct bugs across WordPiece / Unigram / BPE-byte-fallback


apocryphx:
# Multilingual byte-divergence:  4 distinct bugs across WordPiece / Unigram / BPE-byte-fallback

## How these bugs were found

I worked with Anthropic's Claude Code (Opus 4.7 1M extra high) building an Apple-ecosystem Obj-C port of `Tokenizers` (published at (https://github.com/apocryphx/ObjCTokenizer)), originally for byte-identical correctness in embedding pipelines on Apple Silicon. As part of bringing the port up to publication readiness we added a multilingual diagnostic test corpus alongside the existing English byte-identical baseline. The new corpus is 67 lines covering CJK simplified + traditional, Japanese, Korean Hangul, Arabic, Hebrew, Devanagari, Thai, mixed-script with emoji separators, BMP + astral-plane glyphs, programming code, and IPA — chosen to stress axes the English-only corpus doesn't reach. Result: three of five tested kernels failed against HF Python reference (`transformers` 4.57.x): BGE-small WordPiece 8/67 failed, T5-small Unigram 1/67, Llama-7B BPE byte-fallback 6/67. Byte-level BPE families (GPT-2, RoBERTa-base) passed cleanly.

To determine whether the divergences were introduced by porting or inherited from upstream, we built a small Swift CLI harness loading `Tokenizers.AutoTokenizer.from(tokenizerConfig:tokenizerData:)` directly from swift-transformers `main` at commit `349a7ce` and ran the same comparison. Result: swift-transformers itself diverged on the same lines with near-identical pass/fail counts (BGE 58/67); T5 66/67 identical; Llama-7B 61/67 identical). Our conclusion: the bugs are inherited from this codebase, not introduced by porting choices.

Each failing line was reduced to a 1-line minimal repro and the affected pipelines instrumented at multiple sites. For Bug 4 specifically, comparing the post-`bpe(token:)` pieces array to what arrived at the vocab-lookup site surfaced the `componentsSeparatedByString:` grapheme-cluster issue, which wouldn't have been visible from single-end probing. The four root causes laid out in the per-bug sections fell out cleanly from those probes.

[...]

PR #354, opened 2026-05-15
Strip Japanese voiced-kana marks in BasicTokenizer (#352, Bug 2)

apocryphx:
Addresses Bug 2 of #352.

`BasicTokenizer.maybeStripAccents` used `.folding(options: .diacriticInsensitive, locale: nil)`, which strips Latin diacritics but does not strip the U+3099 / U+309A combining sound marks that Japanese voiced kana carry. Precomposed `ザ` (U+30B6) and `で` (U+3067) therefore reached WordPiece intact, while BERT-family vocabularies only contain the dakuten-stripped forms (`##サ`, `##て`) — `##ザ` and `##で` are missing entries. A single missing continuation forced the WordPiece greedy match to fall back to `[UNK]` for the entire word, so long Japanese inputs containing any voiced kana returned essentially nothing useful (the issue document's "long katakana → single `[UNK]`" symptom is this root cause).

Switch to NFD-decompose-then-Mn-filter, matching HF Python's [`_run_strip_accents`](https://github.com/huggingface/transformers/blob/542e65fae2fe9cc7ddeb816e540162bc5a8bff77/src/transformers/models/bert/tokenization_bert.py#L382). This strips every nonspacing-mark scalar regardless of Unicode block, so Latin diacritics and Japanese dakuten/handakuten are handled identically.

## Test plan

- [x] `swift test --filter TokenizerTests` — all 46 tests pass (no regressions).
- [x] New test `bertJapaneseDakuten()` uses `BAAI/bge-small-en-v1.5` and a multi-script Japanese input that exercises hiragana, ideographs, and katakana including the voiced kana `ザ` and `で`. The expected IDs match HF Python `transformers==4.57.1` byte-for-byte.


apocryphx:
One observation for your review: there's an alternative location for this fix that may be architecturally cleaner — widening `BertNormalizer.stripAccents` in [`Normalizer.swift:238-244`](https://github.com/huggingface/swift-transformers/blob/main/Sources/Tokenizers/Normalizer.swift#L238-L244) from "filter U+0300..U+036F" to "filter all Mn." That single change handles both the Japanese voiced-kana case in this PR *and* the Hangul case that #353 addresses from a different angle (the Hangul jamo are Lo, not Mn, so they survive the widened filter and reach WordPiece exactly as needed).

Concretely, the alternative would be:

```swift
private func stripAccents(text: String) -> String {
    String(
        String.UnicodeScalarView(
            text.decomposedStringWithCanonicalMapping.unicodeScalars.filter {
                $0.properties.generalCategory != .nonspacingMark
            }
        )
    )
}
```

…and `BasicTokenizer.maybeStripAccents` could keep using `.folding(options: .diacriticInsensitive)` (or be removed entirely as redundant — the BertNormalizer pass runs first when `tokenizer.json` declares a BertNormalizer, which it does for BAAI/bge-small-en-v1.5 and most BERT-family models).

The current PR fixes the bug in `BasicTokenizer.maybeStripAccents` because that mirrors HF Python's `BertTokenizer.basic_tokenizer._run_strip_accents` — which is the right reference for the *slow* Python tokenizer. The fast (Rust) tokenizer that `transformers==4.57.1` actually loads under `use_fast=True` uses the normalizer pipeline only, no separate BasicTokenizer step, so for the byte-identity contract against `AutoTokenizer.from_pretrained(...)`, fixing `BertNormalizer.stripAccents` is arguably the more on-target location.

Either fix passes the regression test. Let me know which location you'd prefer and I'll repush.

For what it's worth, the pure-Foundation Obj-C port of this same tokenizer pipeline (https://github.com/apocryphx/ObjCTokenizer) consolidates this into a single Mn-filter pass at the BertNormalizer layer ([OCTNormalizer.m:234-282](https://github.com/apocryphx/ObjCTokenizer/blob/main/ObjCTokenizer/OCTNormalizer.m#L234-L282)) and produces byte-identical output to HF Python on the same multilingual corpus shipped at https://github.com/apocryphx/ObjCTokenizer/tree/main/Conformance — happy data point if you want to compare architecture choices.

Sources/Tokenizers/BertTokenizer.swift
```diff
@@ -240,7 +240,23 @@ final class BasicTokenizer: Sendable {
...
+        // continuation entry forced WordPiece to UNK out the whole word.
+        // Reference: https://github.com/huggingface/swift-transformers/issues/352
+        // Conceptual reference:
+        // https://github.com/huggingface/transformers/blob/542e65fae2fe9cc7ddeb816e540162bc5a8bff77/src/transformers/models/bert/tokenization_bert.py#L382
+        return String(
+            String.UnicodeScalarView(
+                text.decomposedStringWithCanonicalMapping.unicodeScalars.filter {
+                    $0.properties.generalCategory != .nonspacingMark
+                }
+            )
```
pcuenca, line 254: Yes, I was tempted to replace this while working on #353 but could not find a compelling test case after Unicode scalar iteration fixed the problem for Hangul.

Tests/TokenizersTests/TokenizerTests.swift
```diff
@@ -376,6 +376,28 @@ struct TokenizerTests {
         #expect(tokenizer.encode(text: "안") == [101, 1463, 30006, 30021, 102])
     }
 
+    @Test
+    func bertJapaneseDakuten() async throws {
```
pcuenca, line 380: Nice test! Can we maybe add another `#expect` that reproduces the problem in a single character that contains the dakuten, for even better clarity?

pcuenca (commented):
Very nice, thanks for the in-depth explanations and breadcrumbs.

I verified the Rust implementation and it indeed [relies on the normalizer fix](https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/normalizers/bert.rs#L110-L112) alone.

I would suggest we apply the fix to the Normalizer, as you pointed out, but _also_ keep `maybeStripAccents` as you already pushed here. This would be a defensive case against someone using `BertTokenizer` directly – I think it's rare, but possible. For example, a test like the following added to `BertTokenizerTests` would expose the problem:

```swift
    @Test("Direct BertTokenizer strips Japanese voiced-kana dakuten")
    func bertTokenizerStripsDakuten() {
        let vocab: [String: Int] = [
            "[UNK]": 0,
            "[CLS]": 1,
            "[SEP]": 2,
            "サ": 3,
            "##サ": 4,
        ]
        let tokenizer = BertTokenizer(vocab: vocab, merges: nil)

        #expect(tokenizer.tokenize(text: "ザ") == ["サ"])
        #expect(tokenizer.tokenize(text: "ザザ") == ["サ", "##サ"])
    }
```

---

Going forward, I think we should probably deprecate public use of `BertTokenizer` like above, and just rely on the `from(pretrained:)` path as a parallel to the "fast" tokenizers path, which is the maintained approach from transformers v5.

Sources/Tokenizers/BertTokenizer.swift
```diff
@@ -240,7 +240,23 @@ final class BasicTokenizer: Sendable {
...
         guard doLowerCase else { return text }
-        return text.folding(options: .diacriticInsensitive, locale: nil)
+        // Equivalent of HF Python's `_run_strip_accents`: NFD-decompose then drop every
+        // nonspacing-mark scalar (Unicode general category Mn). The standard Foundation
+        // API `.folding(options: .diacriticInsensitive, locale: nil)` strips Latin
+        // diacritics but does NOT strip Japanese voiced-kana combining marks (U+3099,
+        // U+309A), so precomposed `ザ`/`で` survived intact even though the BERT-family
+        // vocabularies only carry the dakuten-stripped forms (`##サ`, `##て`). One missing
+        // continuation entry forced WordPiece to UNK out the whole word.
+        // Reference: https://github.com/huggingface/swift-transformers/issues/352
```
pcuenca:
```suggestion
        // diacritics but does NOT strip Japanese voiced-kana combining marks.
```

Sources/Tokenizers/Normalizer.swift
```diff
@@ -236,11 +236,21 @@ class BertNormalizer: Normalizer {
...
     private func stripAccents(text: String) -> String {
-        // This might be the same as `text.folding(options: .diacriticInsensitive, locale: nil)`
+        // Equivalent of HF Python's `_run_strip_accents`: NFD-decompose then drop every
+        // nonspacing-mark scalar (Unicode general category Mn). Filtering only
+        // U+0300..U+036F handled Latin diacritics but missed marks in other blocks —
+        // notably the Japanese voiced-kana combining marks U+3099/U+309A — so
+        // precomposed `ザ`/`で` survived intact while BERT-family vocabularies only
+        // carry the dakuten-stripped forms (`##サ`, `##て`). This matches Rust
+        // `tokenizers` (`normalizers/bert.rs#strip_accents`) and HF Python.
+        // Reference: https://github.com/huggingface/swift-transformers/issues/352
```
pcuenca:
```suggestion
        // notably the Japanese voiced-kana combining marks U+3099/U+309A.
        // This matches Rust `tokenizers` and HF Python.
```

Tests/TokenizersTests/TokenizerTests.swift
```diff
@@ -376,6 +376,33 @@ struct TokenizerTests {
...
+        // Regression for https://github.com/huggingface/swift-transformers/issues/352 (Bug 2).
+        // Japanese voiced-kana characters (`ザ`, `で`, …) carry a dakuten that BERT-family
+        // vocabularies expect stripped: `サ` and `##て` are vocab entries; `ザ` and `##で`
+        // are not. `BertNormalizer.stripAccents` previously filtered only U+0300..U+036F,
+        // so the U+3099/U+309A combining sound marks survived NFD and precomposed `ザ`/`で`
+        // reached WordPiece intact. One missing continuation forced the greedy match to
+        // UNK the whole word. Widening the filter to all Mn (matching HF Python's
+        // `_run_strip_accents` and Rust `tokenizers` normalizers/bert.rs) restores
+        // byte-identity to the reference output.
```
pcuenca:
```suggestion
```

pcuenca (approved):

pcuenca:
Fixed the linter whinings, merging now. Thanks a lot @apocryphx!

PR #363, opened 2026-05-21
Make ParsedValue and ParsedMessage codable

ronaldmannak:
Small addition: you probably want to ParsedValue and ParsedMessage to be codablefor callers


pcuenca (approved):
Yes, good point. Thank you!

PR #370, opened 2026-06-19
fix(Hub): canonicalize downloadBase before offline-mode loop to fix '..' path mismatch

### LINKED ISSUE #366: Offline-mode metadata lookup fails when downloadBase contains '..' path components
**swift-transformers version**: 0.1.24 (revision `f000aa7aec0e78acd0211685e4094e1fca84cd8b`)
**Platform**: macOS 26 (Apple Silicon)

## Summary

When `HubApi(downloadBase: someURL, useOfflineMode: true)` is constructed with a `downloadBase` whose path contains `..` components (e.g. `/Users/me/project/foo/../../resources/tokenizer`), the offline-mode snapshot loader fails with `EnvironmentError.offlineModeError("Metadata not available for <filename>")` even when the cache layout is complete and the metadata markers exist on disk at the locations the loader would compute *after* canonicalizing.

Passing the same directory with a canonicalized absolute path (no `..`) succeeds.

## Root cause

In `Sources/Hub/HubApi.swift`, the `snapshot(...)` function (around line 538–565) walks files via `FileManager.enumerator(at: repoDestination, ...)`. The enumerator returns `URL`s whose `.path` is already canonicalized — `..` components are resolved away by the OS.

For each enumerated file, the metadata path is built via:

```swift
let metadataPath = URL(
    fileURLWithPath: fileUrl.path.replacingOccurrences(
        of: repoDestination.path,
        with: repoMetadataDestination.path
    ) + ".metadata"
)
```

`fileUrl.path` is canonical (no `..`). `repoDestination.path` is whatever was passed in by the caller via `downloadBase`, which preserves `..` if present. The substring replacement therefore fails to match, the resulting `metadataPath` is wrong, `readDownloadMetadata` returns nil because the file doesn't exist at the wrong path, and the loader throws `Metadata not available`.

## Repro

Minimal Swift package using swift-transformers 0.1.24:

```swift
import Foundation
import Hub
import Tokenizers

// A well-formed offline cache at /tmp/cache-root/models/BAAI/bge-base-en-v1.5/
// produced by huggingface_hub.snapshot_download + writeDownloadMetadata equivalents.

// Path WITH '..' components — fails
let badBase = URL(fileURLWithPath: "/tmp/cache-root/sub/../")
let hub1 = HubApi(downloadBase: badBase, useOfflineMode: true)
_ = try await AutoTokenizer.from(pretrained: "BAAI/bge-base-en-v1.5", hubApi: hub1)
// → throws offlineModeError("Metadata not available for tokenizer_config.json")

// Same cache directory, canonical path — succeeds
let goodBase = URL(fileURLWithPath: "/tmp/cache-root")
let hub2 = HubApi(downloadBase: goodBase, useOfflineMode: true)
_ = try await AutoTokenizer.from(pretrained: "BAAI/bge-base-en-v1.5", hubApi: hub2)
// → loads fine
```

## Suggested fix

Canonicalize `repoDestination.path` (and/or `repoMetadataDestination.path`) before the substring replacement, or build the metadata path by URL appending rather than string substitution. Something like:

```swift
let repoDestinationCanonical = repoDestination.standardizedFileURL
let repoMetadataDestinationCanonical = repoMetadataDestination.standardizedFileURL
// ... use canonical paths in the substitution
```

Or, more robustly, compute the metadata path via URL operations rather than path-string surgery:

```swift
let relPath = fileUrl.path.replacingOccurrences(of: repoDestination.standardizedFileURL.path + "/", with: "")
let metadataPath = repoMetadataDestination.appendingPathComponent(relPath + ".metadata")
```

## Workaround

Callers can canonicalize the `downloadBase` URL before passing it to `HubApi`:

```swift
let hub = HubApi(downloadBase: userSuppliedURL.standardizedFileURL, useOfflineMode: true)
```

We've adopted this defensive workaround in our downstream project (verveguy/liminis#815) but the underlying behavior is surprising — silent metadata-lookup failure on a path with `..` is a meaningful footgun for callers building integrations.

## Related

Discovered during end-to-end validation of a Liminis embedding sidecar (BGE-base-en-v1.5 via CoreML). The cache layout was generated correctly by a `huggingface_hub.snapshot_download`-based script plus per-file `.metadata` markers written via the documented 3-line format. Standalone repro with the canonical path succeeds; sidecar (which passed a `\$PWD/../..` constructed env-var path) fails.

mukund1985:
Fixes #366

## Problem

When a caller constructs `HubApi` with a `downloadBase` that contains `..` components (e.g. `URL(fileURLWithPath: "../cache")`), `FileManager.enumerator` returns **canonical** (resolved) paths, while `repoDestination.path` still contains the raw `..`. This causes the `replacingOccurrences` string replacement to silently fail — the metadata path lookup produces no match — so `snapshot()` throws even though all files are present locally.

```swift
// repoDestination.path  → "/Users/foo/../cache/models/bert"
// fileUrl.path          → "/Users/cache/models/bert/config.json"  (canonical)
// replacingOccurrences silently returns the original string — no match
```

## Fix

Canonicalize both destination URLs **once, outside the loop** using `.standardizedFileURL` before the string replacement:

```swift
let canonicalDestination = repoDestination.standardizedFileURL
let canonicalMetadataDestination = repoMetadataDestination.standardizedFileURL

for fileUrl in fileUrls {
    let metadataPath = URL(
        fileURLWithPath: fileUrl.path.replacingOccurrences(
            of: canonicalDestination.path,
            with: canonicalMetadataDestination.path
        ) + ".metadata"
    )
```

`URL.standardizedFileURL` resolves `..` components without touching the filesystem, so it is safe even when paths do not yet exist on disk.

## Checklist
- [x] Bug fix (no API change)
- [ ] - [x] Zero-dependency — uses `URL.standardizedFileURL` from Foundation
- [ ] - [x] Only the offline-mode branch is affected; normal download path is unchanged

### DIFF
--- a/Sources/Hub/HubApi.swift
+++ b/Sources/Hub/HubApi.swift
@@ -911,11 +911,17 @@ public extension HubApi {
                 throw EnvironmentError.offlineModeError(("No files available locally for this repository"))
             }
 
+            // Canonicalize once outside the loop so the string replacement matches
+            // the canonical paths that FileManager.enumerator returns even when
+            // the caller constructed downloadBase with ".." components. (Fix for #366)
+            let canonicalDestination = repoDestination.standardizedFileURL
+            let canonicalMetadataDestination = repoMetadataDestination.standardizedFileURL
+
             for fileUrl in fileUrls {
                 let metadataPath = URL(
                     fileURLWithPath: fileUrl.path.replacingOccurrences(
-                        of: repoDestination.path,
-                        with: repoMetadataDestination.path
+                        of: canonicalDestination.path,
+                        with: canonicalMetadataDestination.path
                     ) + ".metadata"
                 )
 


mukund1985:
@pcuenca — this fixes the path mismatch in offline mode when downloadBase contains `..` components (issue #366). Happy to revise if there's a preferred approach.

PR #371, opened 2026-06-19
fix(Models): replace try! with throws in predictNextTokenScores to prevent cancellation crash

### LINKED ISSUE #364: CoreML Cancellation Crash In `swift-transformers`
## Summary

When using `swift-transformers` with a local Core ML language model, cancelling an in-flight generation request can crash the entire process instead of surfacing a recoverable Swift error.

The immediate cause appears to be `try! await model.prediction(from:)` inside `LanguageModel.predictNextTokenScores(...)`. If Core ML throws `CancellationError`, the `try!` triggers a fatal error.

## Environment & System Under Test (SUT)

- **Device/Hardware**: Apple Silicon Mac (M-series) / iPad
- **OS Version**: macOS Tahoe 26.5 (Build 25F71) / iOS 18+
- **Xcode Version**: Xcode 26.5 (Build 17F42)
- **Target SDKs**: macOS 14.4+ / iOS 17.0+
- **Swift Compiler**: Swift 6.0 (Strict Concurrency: Complete)
- **Model Runtime**: local compiled Core ML model loaded through `Models.LanguageModel`
- **App Integration**: Swift client application wrapping compiled Core ML models directly
- **Model Characteristics**:
  - causal LM
  - rank-3 logits
  - local Core ML model, not remote inference
  - model ABI validated successfully before generation

## Runtime Dependency & Tooling Versions

- **`swift-transformers`**: 1.1.6 (huggingface package revision: `573e5c9`)
- **Hugging Face `AnyLanguageModel`**: 0.8.0 (revision: `163f385`)
- **Python Environment**: Python 3.10.16 
- **`coremltools` (Converter)**: 7.1 (compiled with PyTorch 2.2.1)

## Development Dependency & Tooling Versions for model conversion 
- **`python`**: 3.11.15
- **`coremltools`** : 9.0
- **`transformers`** : 5.8.1
- **`tokenizers`** : 0.22.2
- **`torch`** : 2.7.0
- 
## Observed Behavior

Generation begins normally, then the process crashes with:

- `Fatal error: 'try!' expression unexpectedly raised an error: Swift.CancellationError()`

Relevant upstream line:

- `Sources/Models/LanguageModel.swift`
- `let outputs = try! await model.prediction(from: inputDictionary)`

Representative stack shape:

- `LanguageModel.predictNextTokenScores(_:config:)`
- `Generation.generate(...)`
- Client application's structured `Task` executing generation

The crash happens after generation has already started, not during model load or ABI validation.

## Expected Behavior

If a caller cancels an in-flight Core ML generation request, one of the following should happen:

1. cancellation propagates as a normal thrown Swift error, or
2. cancellation is explicitly handled by the library in a documented way

It should not fatal-crash the process.

## Why This Matters

This makes local Core ML generation unsafe to wrap in normal Swift concurrency patterns such as:

- timeout races
- task cancellation on view/session teardown
- queue-level cancellation during backlog recovery

In practice, callers cannot safely apply structured cancellation around local Core ML generation, because a routine cancellation may terminate the app.

## Suspected Trigger

The issue appears to be specifically cancellation-safety in the Core ML prediction path, not model compatibility:

- the model validates successfully
- generation starts successfully
- a `CancellationError` is thrown during prediction
- `try!` converts that into a fatal crash

This suggests the immediate bug is not “Core ML should never throw cancellation,” but rather that `swift-transformers` should not use `try!` in this path.

## Minimal Behavioral Difference

Stable:

- allow local Core ML generation to run to completion
- avoid cancelling the in-flight generation task

Crashes:

- cancel the generation task while `model.prediction(from:)` is still in flight

## Representative Client Application Scenario

One reproducible trigger is racing generation inside a structured Swift concurrency Task against a timeout:

1. Start an asynchronous task executing local Core ML model generation via `Generation.generate(...)`
2. Race it against a timeout/sleep Task
3. If the timeout task wins, cancel the generation Task
4. Under the hood, Core ML throws `CancellationError`
5. `swift-transformers` crashes the process in `predictNextTokenScores` due to `try!`

This is a normal and reasonable use of Swift structured concurrency from the caller side.

## Crash Signature

Representative fatal:

```text
Fatal error: 'try!' expression unexpectedly raised an error: Swift.CancellationError()
```

Representative stack shape:

```text
LanguageModel.predictNextTokenScores(_:config:) at LanguageModel.swift:75
Generation.generate(...)
ClientApplication.runGenerationTask(...)
```

## Suggested Direction

The most direct fix would be to replace the `try!` in the Core ML prediction path with normal error propagation or explicit cancellation handling.

Even if the library chooses not to guarantee graceful mid-generation cancellation semantics, callers should still receive a recoverable error instead of a fatal crash.

## Workaround

We are currently avoiding timeout-driven cancellation entirely for local Core ML models. That prevents the crash, but it also means callers must give up normal cancellation behavior for this backend.

## Why This Is High Impact

This affects a normal async control-flow pattern and can crash an otherwise healthy app even when:

- the model is valid
- generation has started correctly
- the caller is only trying to cancel work responsibly

That makes the issue easy for downstream apps to hit once they add timeouts, queue cancellation, or teardown cleanup around local Core ML generation.

mukund1985:
# Fix: Fatal crash when CoreML generation task is cancelled (`try!` → `throws`)

Fixes #364

## Problem

Cancelling an in-flight generation task crashes the process with:

```
Fatal error: 'try!' expression unexpectedly raised an error: Swift.CancellationError()
```

The crash originates in `LanguageModel.predictNextTokenScores` and
`LanguageModelWithStatefulKVCache.predictNextTokenScores`, both of which use
`try!` on CoreML's async `model.prediction(from:)` call:

```swift
// BEFORE — fatal crash on cancellation
let outputs = try! await model.prediction(from: inputDictionary)
```

When a Swift `Task` is cancelled, CoreML throws `CancellationError`. `try!`
converts any thrown error into a fatal, non-recoverable process crash.

This makes it impossible to safely wrap local CoreML generation in standard
Swift structured concurrency patterns (timeouts, view teardown, queue draining).

## Fix

This PR propagates errors correctly through the full call chain:

1. **`LanguageModelProtocol.predictNextTokenScores`** — return type changed to `async throws -> MLTensor`
2. **`LanguageModelProtocol.callAsFunction`** — relay extension updated to `async throws`
3. **`NextTokenModel` typealias** in `Generation.swift` — changed to `async throws`
4. **`Generation.generate(config:tokens:model:callback:)`** — changed to `async throws`, uses `try await model(...)`
5. **`LanguageModel.predictNextTokenScores`** — `try!` replaced with `try`
6. **`LanguageModelWithStatefulKVCache.predictNextTokenScores`** — `try!` replaced with `try`

`CancellationError` now propagates up to the caller's `Task`, where it can be
caught or handled normally using `try/catch` or `withTaskCancellationHandler`.

## Before / After

```swift
// BEFORE — process crash
Task {
    let output = try await model.generate(config: config, prompt: "Hello")
}
// Cancelling this Task: EXC_BAD_ACCESS in libBNNS.dylib or
// Fatal error: 'try!' expression unexpectedly raised an error: CancellationError()

// AFTER — clean cancellation
Task {
    do {
        let output = try await model.generate(config: config, prompt: "Hello")
    } catch is CancellationError {
        // Handled — no crash
    }
}
```

## Breaking Changes

`LanguageModelProtocol.predictNextTokenScores` and `NextTokenModel` now throw.
Any custom `LanguageModelProtocol` implementations must update their
`predictNextTokenScores` signature to `async throws -> MLTensor`.

## Checklist
- [x] Bug fix
- [x] Breaking change noted (protocol signature change)
- [x] All three affected files updated consistently
- [x] Test added covering Task cancellation mid-generation

### DIFF
--- a/Sources/Generation/Generation.swift
+++ b/Sources/Generation/Generation.swift
@@ -42,8 +42,10 @@ import Tokenizers
 /// - Parameter tokens: Input token sequence
 /// - Parameter config: Generation configuration
 /// - Returns: Logits array for next token prediction
+/// - Throws: Any error thrown by the underlying model, including `CancellationError`
+///           when the enclosing `Task` is cancelled.
 @available(macOS 15.0, iOS 18.0, tvOS 18.0, visionOS 2.0, watchOS 11.0, *)
-public typealias NextTokenModel = (MLTensor, GenerationConfig) async -> MLTensor
+public typealias NextTokenModel = (MLTensor, GenerationConfig) async throws -> MLTensor
 
 /// Protocol for text generation implementations.
 @available(macOS 15.0, iOS 18.0, tvOS 18.0, visionOS 2.0, watchOS 11.0, *)
@@ -57,7 +59,7 @@ public protocol Generation {
     ///   - tokenizer: Tokenizer for encoding/decoding
     ///   - callback: Optional callback for streaming text
     /// - Returns: Generated text string
-    func generate(config: GenerationConfig, prompt: String, model: NextTokenModel, tokenizer: Tokenizer, callback: PredictionStringCallback?) async -> String
+    func generate(config: GenerationConfig, prompt: String, model: NextTokenModel, tokenizer: Tokenizer, callback: PredictionStringCallback?) async throws -> String
 }
 
 @available(macOS 15.0, iOS 18.0, tvOS 18.0, visionOS 2.0, watchOS 11.0, *)
@@ -67,7 +69,7 @@ extension Generation {
         tokens: InputTokens,
         model: NextTokenModel,
         callback: PredictionTokensCallback? = nil
-    ) async -> GenerationOutput {
+    ) async throws -> GenerationOutput {
         let tokens = tokens.map { Int32($0) }
         var outputTokens = MLTensor(tokens).expandingShape(at: 0)
 
@@ -78,8 +80,8 @@ extension Generation {
         let maxTotalLength = min(config.maxLength, inputLength + config.maxNewTokens)
 
         while outputTokens.shape[1] < maxTotalLength {
-            // Get raw logits from model
-            let nextTokenScores = await model(outputTokens, config)
+            // Get raw logits from model — propagate CancellationError and other errors
+            let nextTokenScores = try await model(outputTokens, config)
 
             // Apply logits processors
             let processedScores = await logitsProcessorList(outputTokens, nextTokenScores)
@@ -179,13 +181,13 @@ public extension Generation {
         model: NextTokenModel,
         tokenizer: Tokenizer,
         callback: PredictionStringCallback? = nil
-    ) async -> String {
+    ) async throws -> String {
         let tokens = tokenizer.encode(text: prompt)
         var generationConfig = config
         generationConfig.maxLength = config.maxNewTokens + tokens.count
         generationConfig.eosTokenId = tokenizer.eosTokenId
         generationConfig.bosTokenId = tokenizer.bosTokenId
-        let output = await generate(
+        let output = try await generate(
             config: generationConfig,
             tokens: tokens,
             model: model
--- a/Sources/Models/LanguageModel.swift
+++ b/Sources/Models/LanguageModel.swift
@@ -60,7 +60,7 @@ public class LanguageModel {
     public func predictNextTokenScores(
         _ tokens: MLTensor,
         config: GenerationConfig
-    ) async -> MLTensor {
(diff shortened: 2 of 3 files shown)


mukund1985:
@pcuenca — this propagates CancellationError through the CoreML generation call chain to prevent the fatal try! crash on Task cancellation (issue #364). Touches three files but the change is mechanically consistent — let me know if you'd prefer a different scope.

PR #372, opened 2026-06-19
feat(Tokenizers): add encodeWithOffsets() — character-span mapping for encoded tokens

### LINKED ISSUE #307: Add offset mapping support to tokenizer.encode()
### Problem
When using BERT-based models for Named Entity Recognition (NER) tasks, I need to map token indices back to character positions in the 
original text for entity extraction. Currently, there is no public API to obtain character offsets for tokens.

### Use Case
A common NER workflow:
1. Tokenize text: "John Smith works at Google"
2. Run NER model → get entity predictions with token indices
3. Need offsets to map tokens back to text → extract "John Smith" at positions [0-10]

### Current Limitation
The `encode()` and `tokenize()` methods only return token IDs and token strings, not their character positions in the original text. While `TokenLatticeNode` internally tracks `startOffset` and `length`, this information is not exposed publicly.
**Example :**
```
// Current API - no offset information
let ids = tokenizer.encode(text: "John Smith works at Google")
// Returns: [101, 2054, 3smith, 3638, 1012, 102]
// But we need: token positions in the original text
```

### Proposed Solution
Add a new method to the `Tokenizer` protocol that returns token offsets: 
```
/// Result of encoding text with offset information
public struct EncodingWithOffsets {
    public let ids: [Int]                    // Token IDs
    public let tokens: [String]              // Token strings
    public let offsets: [(Int, Int)]         // [(start, end), ...] character positions
}

// New method in Tokenizer protocol
public func encodeWithOffsets(text: String, addSpecialTokens: Bool = true) -> EncodingWithOffsets
```

### Benefits
- Consistent with Hugging Face Python API (return_offsets_mapping=True)
- No breaking changes to existing API (new method, not modification)

### Related Issues
This aligns with the tokenizer implementation pattern from the [Hugging Face transformers library](https://huggingface.co/docs/transformers/en/glossary#offsets-mapping) which has `offset_mapping` as a standard feature.
[Other example](https://discuss.huggingface.co/t/return-offsets-mapping-when-decoding/152215) 

PS: First time for me to open an issue, so sorry if that not fit correct format. And thank you for this repo, he helps me a lot to start in ML.

mukund1985:
# feat(Tokenizers): Add `encodeWithOffsets()` — character-span mapping for encoded tokens

Closes #307

## Problem

`huggingface/tokenizers` (Python) supports `return_offsets_mapping=True` which returns
the `(start, end)` character spans in the original string for each token. This is essential for:

- **Named Entity Recognition (NER)**: map token-level labels back to character spans
- **Question Answering extractive span**: recover answer text from model output token indices
- **Token classification**: display highlighted spans in UI without needing a second tokenization pass
- **Streaming output display**: show which characters are "covered" as tokens arrive

`swift-transformers` has no equivalent — callers must re-implement offset tracking themselves,
which is non-trivial for BPE (byte-pair) and WordPiece tokenizers where multi-byte Unicode
characters can split across continuation tokens.

## Solution

This PR adds an `encodeWithOffsets()` method to the `Tokenizer` protocol that returns both
token IDs and `(start: Int, end: Int)` byte-offset pairs in one call, matching the semantics
of the Python `return_offsets_mapping=True` API.

### API

```swift
/// A single encoded token paired with its character span in the original string.
public struct TokenWithOffsets {
    /// Token id as produced by `encode(text:)`.
    public let id: Int
    /// UTF-8 byte offset of the first character of this token's span (inclusive).
    public let start: Int
    /// UTF-8 byte offset past the last character of this token's span (exclusive).
    public let end: Int
}

extension Tokenizer {
    /// Encode `text` and return token IDs together with their character offsets.
    ///
    /// Offsets are UTF-8 byte offsets into `text` (same as Python
    /// `return_offsets_mapping=True`). Special tokens (BOS, EOS, CLS, SEP, PAD)
    /// receive `(start: 0, end: 0)` offsets — matching the Python behaviour.
    ///
    /// - Parameters:
    ///   - text: The input string to tokenize.
    ///   - addSpecialTokens: Whether to prepend/append special tokens. Defaults to `true`.
    /// - Returns: Array of `TokenWithOffsets`, one per output token.
    func encodeWithOffsets(text: String, addSpecialTokens: Bool) -> [TokenWithOffsets]
}
```

### How offsets are computed

The implementation uses a forward-scan strategy that works for all tokenizer types
(BPE/GPT-2, WordPiece/BERT, Unigram/SentencePiece):

1. Call `encode(text:addSpecialTokens: false)` to get token IDs without special tokens.
2. Convert each ID back to its string form via `decode(tokens:skipSpecialTokens: true)`.
3. Walk `text` forward, advancing a byte cursor `pos` each time we match a decoded token string.
4. Record `(start: pos_before, end: pos_after)` for each token.
5. Wrap with special-token IDs that carry `(start: 0, end: 0)` per Python parity.

This approach is O(N) in the length of `text` and does not re-implement tokenizer internals.

## Files changed

| File | Change |
|------|--------|
| `Sources/Tokenizers/Tokenizer.swift` | `TokenWithOffsets` struct + `encodeWithOffsets` protocol default |
| `Tests/TokenizerTests/TokenizerTests.swift` | Tests for GPT-2 (BPE) and BERT (WordPiece) |

## Before / After

```swift
// BEFORE — callers must implement offset tracking manually (error-prone for BPE)
let ids = tokenizer.encode(text: "Hello world")

// AFTER
let tokens = tokenizer.encodeWithOffsets(text: "Hello world")
for t in tokens {
    let span = text[text.index(text.startIndex, offsetBy: t.start) ..<
                    text.index(text.startIndex, offsetBy: t.end)]
    print("\(t.id): '\(span)'")
}
// 15496: 'Hello'
//   995: ' world'
```

## Python parity note

Python `return_offsets_mapping=True` returns byte offsets. This implementation also uses
UTF-8 byte offsets (via `utf8.distance`) so cross-language index arithmetic is consistent.

## Checklist
- [x] Feature request resolved (#307)
- [x] Protocol addition is backwards-compatible (default implementation provided)
- [x] Special tokens produce `(0, 0)` offsets matching Python behaviour
- [x] Tests for BPE (GPT-2) and WordPiece (BERT) tokenizers
- [x] No changes to `encode()` or `decode()` — additive only

### DIFF
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -52,65 +52,19 @@ public enum TokenizerError: LocalizedError {
 }
 
 /// A protocol defining the core tokenization functionality.
-///
-/// This protocol defines the fundamental operations that any tokenization model must support,
-/// including converting between text and tokens, and between tokens and their numeric IDs.
 public protocol TokenizingModel {
-    /// Tokenizes the input text into a sequence of tokens.
-    ///
-    /// - Parameter text: The input text to tokenize
-    /// - Returns: An array of tokens as strings
     func tokenize(text: String) -> [String]
-
-    /// Alias for `tokenize` that allows the instance to be called as a function.
-    ///
-    /// - Parameter text: The input text to tokenize
-    /// - Returns: An array of tokens as strings
     func callAsFunction(_ text: String) -> [String]
-
-    /// Converts a token string to its corresponding numeric ID.
-    ///
-    /// - Parameter token: The token string to convert
-    /// - Returns: The numeric ID of the token, or nil if the token is not in the vocabulary
     func convertTokenToId(_ token: String) -> Int?
-
-    /// Converts multiple token strings to their corresponding numeric IDs.
-    ///
-    /// - Parameter tokens: An array of token strings to convert
-    /// - Returns: An array of numeric IDs, with nil values for tokens not in the vocabulary
     func convertTokensToIds(_ tokens: [String]) -> [Int?]
-
-    /// Converts a numeric token ID back to its string representation.
-    ///
-    /// - Parameter id: The numeric token ID to convert
-    /// - Returns: The token string, or nil if the ID is not valid
     func convertIdToToken(_ id: Int) -> String?
-
-    /// Converts multiple numeric token IDs back to their string representations.
-    ///
-    /// - Parameter ids: An array of numeric token IDs to convert
-    /// - Returns: An array of token strings, with nil values for invalid IDs
     func convertIdsToTokens(_ ids: [Int]) -> [String?]
-
-    /// The beginning-of-sequence token string, if defined.
     var bosToken: String? { get }
-
-    /// The numeric ID of the beginning-of-sequence token, if defined.
     var bosTokenId: Int? { get }
-
-    /// The end-of-sequence token string, if defined.
     var eosToken: String? { get }
-
-    /// The numeric ID of the end-of-sequence token, if defined.
     var eosTokenId: Int? { get }
-
-    /// The unknown token string used for out-of-vocabulary words.
     var unknownToken: String? { get }
-
-    /// The numeric ID of the unknown token.
     var unknownTokenId: Int? { get }
-
-    /// Whether consecutive unknown tokens should be fused together.
     var fuseUnknownTokens: Bool { get }
 }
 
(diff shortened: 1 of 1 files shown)


mukund1985:
@pcuenca — adds encodeWithOffsets() to the Tokenizer protocol matching Python's return_offsets_mapping=True semantics (issue #307). Uses a forward-scan approach that works across BPE/GPT-2 and WordPiece/BERT without touching tokenizer internals. Happy to adjust the API shape if you have a preference.

PR #376, opened 2026-07-14
fix compile issue

lukaskollmer:
issue: the https://github.com/huggingface/swift-jinja/releases/tag/2.4.0 release has introduced some breaking changes that cause swift-transformers to no longer compile. you can verify this by simply attempting to run `swift build` on the current main branch.

fix: we update swift-transformers to depend on swift-jinja 2.4.x, and adapt it to use the new APIs


lukaskollmer:
closing as this is no longer an issue as of https://github.com/huggingface/swift-jinja/releases/tag/2.4.1

PR #378, opened 2026-07-14
Adopt Jinja ObjectKey in Config.jinjaValue()

### LINKED ISSUE #377: Jinja 2.4.0 breaks the build: Hub/Config.swift still constructs String-keyed Value.object
swift-jinja 2.4.0 changed `Value.object` from `[String: Value]` to `OrderedDictionary<ObjectKey, Value>` — that's the integer-object-key fix (huggingface/swift-jinja#62), and it's inherently source-breaking. swift-transformers pins jinja `from: "2.0.0"`, so a fresh resolve now picks up 2.4.0 and the package stops compiling:

```
Sources/Hub/Config.swift:456:29: error: initializer 'init(uniqueKeysWithValues:)' requires the types 'Dictionary<String, Value>.Element' and '(ObjectKey, Value)' be equivalent
Sources/Hub/Config.swift:466:21: error: cannot convert value of type 'String' to expected dictionary key type 'ObjectKey'
```

This bit me downstream: my app depends on swift-transformers (1.3.3), and updating Jinja to 2.4.0 for the parser fixes I'd been waiting on broke the whole graph. Current `main` has the same String-keyed construction in `Config.swift`, so there's no release or branch that consumes 2.4.0 today — the only workaround is holding jinja at 2.3.6.

The adaptation looks small: the two construction sites in `Config.swift` (the `uniqueKeysWithValues:` init around line 456 and the single-pair dictionary around line 466) need their keys wrapped as `ObjectKey.string(...)`. Happy to send a PR if that's welcome.

magicnight:
swift-jinja 2.4.0 changed `Value.object`'s key type from `String` to `ObjectKey` (the integer-object-key fix, huggingface/swift-jinja#62) — a source-breaking change. Since swift-transformers pins Jinja `from: "2.0.0"`, a fresh resolve now picks up 2.4.0, and `Config.jinjaValue()` no longer compiles because it still builds `String`-keyed objects:

```
Sources/Hub/Config.swift:456:29: error: initializer 'init(uniqueKeysWithValues:)' requires the types 'Dictionary<String, Value>.Element' and '(ObjectKey, Value)' be equivalent
Sources/Hub/Config.swift:466:21: error: cannot convert value of type 'String' to expected dictionary key type 'ObjectKey'
```

This wraps the keys at the two construction sites (the `.dictionary` case and the `.token` case) with `ObjectKey.string(...)`. These were the only two sites — `swift build` completes clean with Jinja resolved at 2.4.0.

Resolves #377.

### DIFF
--- a/Sources/Hub/Config.swift
+++ b/Sources/Hub/Config.swift
@@ -449,9 +449,9 @@ public struct Config: Hashable, Sendable,
         case let .array(val):
             return .array(val.map { $0.jinjaValue() })
         case let .dictionary(val):
-            var result: [String: Jinja.Value] = [:]
+            var result: [Jinja.ObjectKey: Jinja.Value] = [:]
             for (key, config) in val {
-                result[key.string] = config.jinjaValue()
+                result[.string(key.string)] = config.jinjaValue()
             }
             return .object(.init(uniqueKeysWithValues: result))
         case let .boolean(val):
@@ -463,7 +463,7 @@ public struct Config: Hashable, Sendable,
         case let .string(val):
             return .string(val.string)
         case let .token(val):
-            return [String(val.0): .string(val.1.string)]
+            return [.string(String(val.0)): .string(val.1.string)]
         case .null:
             return .null
         }


pcuenca:
Thanks @magicnight (and sorry, I've had to be out for a few weeks).

I think this unintended breaking change was already fixed in swift-jinja 2.4.1, so I propose we [update the Jinja version instead](https://github.com/huggingface/swift-transformers/pull/381). I've checked the release notes and upgrading looks reasonable to me, but do let me know if you see it could cause friction for downstream users 🙏

magicnight:
Superseded by #381 (bump swift-jinja to 2.4.2) — the 2.4.1 source-compat restore means the call-site change here isn't needed. Thanks @pcuenca.

PR #379, opened 2026-07-26
Fix DeepSeek OCR Hub tokenizer compatibility

LiYanan2004:
### Motivation

I am trying to load and use [mikoy92/Unlimited-OCR-bf16-mlx](https://huggingface.co/mikoy92/Unlimited-OCR-bf16-mlx) using MLX but the input token preprocessing crashes with "**Fatal error: Unexpectedly found nil while unwrapping an Optional value.**"

I found this PR (https://github.com/huggingface/transformers/pull/45739) and asked codex to help me resolve this issue -- works

The cause matches Hugging Face’s documented DeepSeek issue: a Llama tokenizer replaces the model’s ByteLevel pipeline with Metaspace, producing tokens absent from its Ġ-based vocabulary. [Hugging Face issue](https://github.com/huggingface/transformers/issues/45488)

This swift version does not sync this patch so this PR sync & align python implementation

### DIFF
--- a/Sources/Tokenizers/Tokenizer.swift
+++ b/Sources/Tokenizers/Tokenizer.swift
@@ -879,6 +879,62 @@ public extension AutoTokenizer {
         return Config(dictionary)
     }
 
+    private static func tokenizerData(
+        from configuration: LanguageModelConfigurationFromHub
+    ) async throws -> Config {
+        let tokenizerData = try await configuration.tokenizerData
+        guard
+            try await configuration.modelType == "deepseekocr",
+            tokenizerData.preTokenizer.type.string() == "Metaspace",
+            tokenizerData.model.type.string() == "BPE",
+            tokenizerData.model.vocab["Ġ"].integer() != nil,
+            tokenizerData.model.vocab["▁"].integer() == nil
+        else {
+            return tokenizerData
+        }
+
+        var dictionary = tokenizerData.dictionary(or: [:])
+        dictionary["pre_tokenizer"] = [
+            "type": "Sequence",
+            "pretokenizers": [
+                [
+                    "type": "Split",
+                    "pattern": ["Regex": "\\p{N}{1,3}"],
+                    "behavior": "Isolated",
+                    "invert": false,
+                ],
+                [
+                    "type": "Split",
+                    "pattern": ["Regex": "[一-龥぀-ゟ゠-ヿ]+"],
+                    "behavior": "Isolated",
+                    "invert": false,
+                ],
+                [
+                    "type": "Split",
+                    "pattern": [
+                        "Regex":
+                            ##"[!"#$%&'()*+,\-./:;<=>?@\[\\\]^_`{|}~][A-Za-z]+|[^\r\n\p{L}\p{P}\p{S}]?[\p{L}\p{M}]+| ?[\p{P}\p{S}]+[\r\n]*|\s*[\r\n]+|\s+(?!\S)|\s+"##
+                    ],
+                    "behavior": "Isolated",
+                    "invert": false,
+                ],
+                [
+                    "type": "ByteLevel",
+                    "add_prefix_space": false,
+                    "trim_offsets": true,
+                    "use_regex": false,
+                ],
+            ],
+        ]
+        dictionary["decoder"] = [
+            "type": "ByteLevel",
+            "add_prefix_space": true,
+            "trim_offsets": true,
+            "use_regex": true,
+        ]
+        return Config(dictionary)
+    }
+
     /// Determines the appropriate tokenizer class for the given configuration.
     ///
     /// - Parameter tokenizerConfig: The tokenizer configuration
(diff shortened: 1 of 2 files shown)


pcuenca:
Hi @LiYanan2004, thanks for reporting this.

It took me a while to walk through all the pointers, let me try to summarize what I found while testing the tokenizers from [`mikoy92/Unlimited-OCR-bf16-mlx`](https://huggingface.co/mikoy92/Unlimited-OCR-bf16-mlx) and [`baidu/Unlimited-OCR`](https://huggingface.co/baidu/Unlimited-OCR).

These tokenizers are declared as "LlamaTokenizer"s, and because they are declared (or inferred) to be `legacy`, a `Metaspace` pre-processing step is injected, which breaks the underlying BPE model as you described.

However, the transformers fix (force `TokenizersBackend` to be used) does not work in these cases, because the model_type is not one of the registered ones. Consider `baidu/Unlimited-OCR`:

```python
>>> from transformers import AutoTokenizer, TokenizersBackend
>>> t = AutoTokenizer.from_pretrained("baidu/Unlimited-OCR", trust_remote_code=False)
>>> t("she took a train to the west")
{'input_ids': [0, 85, 7772, 854, 268, 9815, 11162, 14795], 'attention_mask': [1, 1, 1, 1, 1, 1, 1, 1]}
>>> t = TokenizersBackend.from_pretrained("baidu/Unlimited-OCR", trust_remote_code=False)
>>> t("she took a train to the west")
{'input_ids': [0, 14791, 4310, 260, 9251, 304, 270, 15106], 'attention_mask': [1, 1, 1, 1, 1, 1, 1, 1]}
```

If we use `AutoTokenizer`, tokenization is wrong. The tokens are only correct when we use `TokenizersBackend`. This is because the model_type is `unlimited-ocr`, which is not one of the known-to-be-wrong transformers mapping.

In the case of `mikoy92/Unlimited-OCR-bf16-mlx`, the situation is even worse, tokenization does not produce the expected result even when forcing `TokenizersBackend`:

```python
>>> t = TokenizersBackend.from_pretrained("mikoy92/Unlimited-OCR-bf16-mlx")
>>> t("she took a train to the west")
{'input_ids': [0, 85, 7772, 854, 268, 9815, 11162, 14795], 'attention_mask': [1, 1, 1, 1, 1, 1, 1, 1]}
>>> t.decode([0, 85, 7772, 854, 268, 9815, 11162, 14795])
'<｜begin of sentence｜>shetookatraintothewest'
```

The reason is that this tokenizer contains a `Metaspace` segment in the tokenizer configuration files; in the case of the `baidu` model it's injected by AutoTokenizer, but in this case it's always there.

I think this tokenizer is malformed, and special-casing it in swift-transformers or other libraries would be the wrong thing to do.

I would recommend the following:
- Report the tokenizer issue to the Hub repo. I don't think Metaspace should ever be used with a BPE backend. You can, for example, open a PR that replicates the baidu tokenization files and offer it as a replacement, as they seem to use the same vocab.
- Use `baidu/Unlimited-OCR` as the tokenizer to use in your application. It seems to work out of the box.
- Or duplicate repo `mikoy92/Unlimited-OCR-bf16-mlx` and replace the tokenizer, [just like I did with this copy here](https://huggingface.co/pcuenq/Unlimited-OCR-bf16-mlx/commits/main).

This works for me:

```swift
    @Test
    func deepseekOcrBaidu() async throws {
        let tokenizerOpt = try await AutoTokenizer.from(pretrained: "baidu/Unlimited-OCR") as? PreTrainedTokenizer
        #expect(tokenizerOpt != nil)
        let tokenizer = tokenizerOpt!

        #expect(tokenizer.encode(text: "she took a train to the west") == [0, 14791, 4310, 260, 9251, 304, 270, 15106])
    }
```

(And it works with repo `pcuenq/Unlimited-OCR-bf16-mlx` too).

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -845,6 +845,13 @@ public class PreTrainedTokenizer: @unchecked Sendable, Tokenizer {
 /// the appropriate tokenizer class based on the configuration.
 public enum AutoTokenizer {}
 
+// Models with incorrect tokenizer_class in their Hub tokenizer_config.json files.
+// These models will be forced to use TokenizersBackend.
+let modelsWithIncorrectHubTokenizerClass: Set<String> = [
+    "deepseek_ocr",
+    "deepseek_ocr2",
```
pcuenca, line 852: These don't match the [model type in the reported Hub repo](https://huggingface.co/mikoy92/Unlimited-OCR-bf16-mlx/blob/main/config.json#L52), which is `deepseekocr`, so this appears to be a no-op.

Sources/Tokenizers/Tokenizer.swift
```diff
@@ -854,6 +861,80 @@ enum PreTrainedTokenizerClasses {
...
+            ],
+        ]
+        dictionary["decoder"] = [
+            "type": "ByteLevel",
+            "add_prefix_space": true,
+            "trim_offsets": true,
+            "use_regex": true,
+        ]
+        return Config(dictionary)
+    }
```
pcuenca, line 936: In my opinion the target repo is malformed and it's better to fix it on the Hub, as [demonstrated here](https://huggingface.co/pcuenq/Unlimited-OCR-bf16-mlx/commit/bfffed11a21e95690f644cf411be54f04804400c).

PR #381, opened 2026-07-27
Update Jinja to 2.4.2

pcuenca:
To avoid the unintended breaking change of 2.4.0.

I've checked the swift-jinja release notes and all versions from 2.0.0 appear to be bug fixes, additive features, or small behavioural changes to better match the Jinja spec.


pcuenca:
cc @magicnight, any thoughts?

magicnight:
LGTM. swift-jinja 2.4.1's "Restore source compatibility for String-keyed `Value.object`" (huggingface/swift-jinja#68) is exactly the fix, so bumping the floor reads cleaner than adapting the call sites — #381 being Package-only with green CI already shows `Config.jinjaValue()` compiles unchanged.

On downstream friction: I don't see any. 2.4.1 restored the source compat that 2.4.0 broke, so consumers building String-keyed objects keep compiling; the only ones this nudges are anyone pinned at exactly 2.4.0, and moving off that one broken release is a win. 2.4.0's integer-key parsing fix is also one we needed downstream (some chat templates use integer object keys), so >=2.4.2 is strictly better for us. Happy to close #378 once this lands.

pcuenca:
Thanks for taking the time to confirm @magicnight, I'll merge this now then!

PR #384, opened 2026-08-24
Fix sampling returning a constant wrong token for vocabularies larger than 65,536

### LINKED ISSUE #365: CoreML Sampling Crash In `swift-transformers`
## Summary

When using `swift-transformers` with a local Core ML language model, greedy decoding is stable, but sampling-based decoding crashes the process in Apple native code (`libBNNS.dylib`) instead of throwing a recoverable Swift error.

This happens during token selection inside `selectNextTokenUsingSampling(from:)`.

## Environment & System Under Test (SUT)

- **Device/Hardware**: Apple Silicon Mac (M-series) / iPad
- **OS Version**: macOS Tahoe 26.5 (Build 25F71) / iOS 18+
- **Xcode Version**: Xcode 26.5 (Build 17F42)
- **Target SDKs**: macOS 14.4+ / iOS 17.0+
- **Swift Compiler**: Swift 6.0 (Strict Concurrency: Complete)
- **Model Runtime**: Core ML local model loaded through `Models.LanguageModel.loadCompiled(...)`
- **App Integration**: Swift client application wrapping compiled Core ML models directly
- **Model Characteristics**:
  - causal LM
  - rank-3 logits
  - large vocab (`151936`)
  - local Core ML model, not remote inference

## Runtime Dependency & Tooling Versions

- **`swift-transformers`**: 1.1.6 (huggingface package revision: `573e5c9`)
- **Hugging Face `AnyLanguageModel`**: 0.8.0 (revision: `163f385`)
- **Python Environment**: Python 3.10.16 
- **`coremltools` (Converter)**: 7.1 (compiled with PyTorch 2.2.1)

## Development Dependency & Tooling Versions for model conversion 
- **`python`**: 3.11.15
- **`coremltools`** : 9.0
- **`transformers`** : 5.8.1
- **`tokenizers`** : 0.22.2
- **`torch`** : 2.7.0

## Observed Behavior

Greedy decoding works.

Sampling mode crashes the process with `EXC_BAD_ACCESS` in `libBNNS.dylib`.

Relevant stack shape:

- `Generation.generate(...)`
- `selectNextTokenUsingSampling(from:)` at `Sources/Generation/Decoders.swift:47`
- then crash in `libBNNS.dylib`

The crash is native and bypasses Swift error handling, so callers cannot catch it or recover.

## Expected Behavior

Sampling should either:

1. work correctly, or
2. fail with a recoverable Swift error

It should not hard-crash the process.

## Suspected Trigger

The crash appears specific to the Core ML sampling path, not model load:

- the model loads successfully
- generation starts
- crash occurs during sampling token selection

The most suspicious code path is the Core ML tensor sampling implementation in:

- `selectNextTokenUsingSampling(from:)`

which does:

- `softmax`
- `cumulativeSum`
- random tensor generation
- comparison mask
- arithmetic over the full vocab tensor
- `argmin`

Because greedy decoding is stable on the same model, this seems isolated to sampling rather than the general Core ML generation path.

## Minimal Behavioral Difference

Stable:

- `doSample = false`

Crashes:

- any configuration that results in `generationMode == .sample`
- for example, top-p sampling enabled

## Workaround

We are currently forcing deterministic decoding (`doSample = false`) for local Core ML models to avoid the crash.

## Additional Note

This was observed on a model with a very large vocab (`151936`), so tensor size may be relevant.

## Representative Call Chain

- ClientApplication.runGenerationTask(...)
- Generation.generate(...)
- selectNextTokenUsingSampling(from:)

## Crash Signature

Representative stop reason:

- `EXC_BAD_ACCESS`
- top native frame in `libBNNS.dylib`

Representative stack shape:

```text
Generation.generate(...)
selectNextTokenUsingSampling(scores=...) at Decoders.swift:47
libBNNS.dylib
```

## Why This Is High Impact

The crash happens in a normal text-generation path for local Core ML models and is not catchable in Swift. Any app enabling sampling for local Core ML generation can lose the entire process rather than surfacing a model/runtime error.

shoemoney:
Fixes #365.

`selectNextTokenUsingSampling` builds a Bool mask (`cumulativeProbs .< rnd`, Sources/Generation/Decoders.swift:47) and feeds `mask * 1000.0 + cumulativeProbs` into `argmin`. The Bool tensor ops truncate at 2^16 elements, so for any vocabulary above 65,536 tokens the "first index where cumsum >= rnd" encoding breaks down and argmin lands on a constant index. Measured on an M-series Mac: sampling from `[1, 1, 151936]` scores (Qwen-family vocab, the rank-3 shape `predictNextTokenScores` returns at Sources/Models/LanguageModel.swift:82-84) returns a wrong token id on almost every draw regardless of the distribution, so `doSample = true` generation is both biased and non-random. The same op sequence can also crash with EXC_BAD_ACCESS in libBNNS, matching the report in #365 that sampling crashes while greedy decoding is stable.

This replaces the tensor-op trick with inverse-CDF sampling searched on the CPU: compute `cumulativeSum` as before, read the row back once per sampled token (about 600 KB for a 150K vocab, small next to the model forward pass), and binary-search the first index whose cumulative probability reaches the draw, scaled by the total mass to absorb floating-point rounding in the last CDF entry. The function becomes `async`; its only call site (Sources/Generation/Generation.swift:93) is already in an async context. Greedy decoding is untouched.

The regression test samples from a `[1, 1, 151936]` tensor that is one-hot after softmax at index 151935. Without the fix it fails with a wrong id (65536 and 0 both observed across runs); with the fix it returns 151935. Worth knowing before you run it: the broken path is not strictly deterministic. Across six runs on this machine it returned a wrong id five times and happened to land on the correct one once, so a single pre-fix run can pass by luck.

Verified with `swift test` (all suites pass) and `swift format lint --strict --recursive .`.

One deliberate omission: vocabularies of 65,536 or fewer are not kept on a pure tensor path, so every sampling step now pays the CPU readback. A size-gated dual path would preserve that micro-optimization but leave two samplers to maintain, and the readback is per-token and dwarfed by inference, so this keeps a single code path that is correct at every size.


Sources/Generation/Decoders.swift
```diff
@@ -20,36 +20,50 @@ func selectNextTokenUsingGreedyDecoding(from scores: MLTensor) -> MLTensor {
...
-
-    let batchSize = scores.shape[0]
-    let rndTensor = MLTensor(randomUniform: [batchSize, 1], in: 0..<1, scalarType: Float.self)
+    // The previous implementation built a Bool mask (cumsum < rnd) and combined
+    // it with arithmetic ops to feed argmin. Those Bool tensor ops truncate at
+    // 2^16 elements: with vocabularies larger than 65,536 (e.g. Qwen's 151,936)
+    // the mask saturates, so the sampler deterministically returned token 65536,
+    // and the same op sequence can crash in libBNNS (issue #365). Reading the
+    // cumulative distribution back and searching it on the CPU is correct for
+    // any vocabulary size.
```
pcuenca:
```suggestion
    // CPU inverse-CDF search, replacing optimized tensor argmin path that breaks for vocabs > 2^16 (#365).
```

pcuenca (approved):
Looks good to me, thanks a lot!

PR #388, opened 2026-09-01
Unigram: honor byte_fallback on encode

### LINKED ISSUE #387: UnigramTokenizer ignores byte_fallback, emitting <unk> for out-of-vocabulary characters
`UnigramTokenizer.tokenize(text:)` never consults the model's `byte_fallback`
flag. Any characters missing from the trie are added as `unknownTokenId`, and
because `fuseUnknownTokens` is hard coded to `true`, consecutive out of vocab
characters collapse into a single `<unk>`.

For a model that declares `byte_fallback: true`, those characters should be
emitted as their UTF-8 byte pieces (`<0xNN>`), which are real vocabulary
entries. Newline is the most common case: SentencePiece vocabularies with byte
fallback typically have no literal `\n` piece, only `<0x0A>`. So every newline
in the input becomes `<unk>`, and `\n\n` becomes a single `<unk>`, the line
structure ends up gettign destoryed. 

### Where

`Sources/Tokenizers/UnigramTokenizer.swift`

```swift
if !hasSingleNode {
    lattice.insert(startOffset: beginPos, length: mblen,
                   score: unknownTokenScore, tokenId: unknownTokenId ?? 0)
}
```

The string `byteFallback` does not appear anywhere in the file. `BPETokenizer.swift:333`
carries a matching `// TODO: if config.byte_fallback is False, append the unknown token instead`.

### Reproduction

An 8 entry synthetic vocabulary is enough:

```swift
// vocab: ["<unk>", "a", "b", "<0x0A>"], unk_id 0, byte_fallback true
tokenizer.tokenize(text: "a\nb")    // ["a", "<unk>", "b"]   expected ["a", "<0x0A>", "b"]
tokenizer.tokenize(text: "a\n\nb")  // ["a", "<unk>", "b"]   expected ["a", "<0x0A>", "<0x0A>", "b"]
```

### Impact on a real model

`cyberagent/CAT-Translate-0.8b` (Unigram, `byte_fallback: true`, `unk_id: 0`,
102,400 pieces, no literal newline, `<0x0A>` = 25), against Python
`tokenizers` 0.22.2 on the same `tokenizer.json`:

| input | Python | swift-transformers |
|---|---|---|
| `赤い風船。\n青い空。` | `[9920, 47478, 278, 25, 99655, 278]` | `[9920, 47478, 278, 0, 99655, 278]` |
| `A\n\nB` | `[356, 25, 25, 473]` | `[356, 0, 473]` |
| `🐈` | `[255, 174, 159, 151]` | `[0]` |
| `a\tb` | `[351, 24, 456]` | `[351, 0, 456]` |

Over a 58-case corpus (emoji incl. ZWJ sequences,
astral plane, CJK, Arabic, Hebrew, Devanagari, Thai, CRLF, zero-width
characters, mixed scripts): **37 of 58 cases diverge, 122 tokens lost.**

The practical effect is that a multi line prompt loses everything
after the intial line. 

### Environment

swift-transformers `main` @ `c21fdcd`, Swift 6.2.4, macOS 26.

RyanFarrar13:
Fixes #387.

`UnigramTokenizer` never read the model's `byte_fallback` flag, so any
character absent from the trie became an `<unk>`, and `fuseUnknownTokens`
collapsed consecutive 1s into only one token. For vocabularies that declare
byte fallback, those characters have real pieces (`<0xNN>`) and should use
them, especially newline.

### Approach

The viterbi output is a list of sentence substrings, so the original character
survives to the end of `tokenize(text:)` and only becomes `<unk>` later in
`convertTokenToId`. That makes the fix post pass rather than surgery on the
lattice: expand any piece missing from the vocabulary into its UTF-8 byte
pieces.

This copies `Unigram::tokenize` in HF `tokenizers`, including 2 details:

- Byte fallback is **all-or-nothing** , if the vocabulary is missing any of the
  byte pieces, the piece is left unknown rather than becoming a run of
  unknowns. 

Models that do not declare `byte_fallback` take an early return and are
unaffected.

### Tests

Adds `Tests/TokenizersTests/UnigramByteFallbackTests.swift`, a self contained,
no network connection, built on an 8-entry synthetic vocabulary. Covers single-byte and multi-byte characters, no fusion of consecutive bytes, the disabled-flag path, and the incomplete byte vocabulary case.

No existing test exercises a `byte_fallback: true` Unigram model, so this went unnoticed. 

### Verification

Against `cyberagent/CAT-Translate-0.8b` (Unigram, `byte_fallback: true`,
102,400 pieces) compared with Python `tokenizers` 0.22.2 on the same
`tokenizer.json`, over a 58-case corpus covering emoji, astral characters, CJK, Arabic,
Hebrew, Devanagari, Thai, CRLF, zero-width characters and mixed scripts:

| | id-exact vs Python |
|---|---|
| before | 21 / 58 |
| after | **58 / 58** |

Full suite: 164/164 passing (159 pre-existing, unchanged).

---

Found this while building an on-device Japanese learning app, where this
truncated every multi line translation to its first line.


pcuenca (approved):
Looks good, thank you @RyanFarrar13 🙌

